PR #3, opened 2021-09-09
Entity filter

krebsni:
Entity filter functionality complete, various use cases tested and working.


buehlefs (changes requested):

plugins/entity_filter.py
```diff
@@ -0,0 +1,432 @@
...
+        raise ValueError("Attribute setting not specified!")
+    
+    # Filter rows: get requested entities from input file 
+    output_entities_id_list = []
+    # Random sampling as in https://math.stackexchange.com/questions/846036/can-i-uniformly-sample-from-n-distinct-elements-where-n-is-unknown-but-fini
+    output_entities_random_rows = []
+    sampling_counter = 0
+    mimetype = mimetypes.MimeTypes().guess_type(input_file_url)[0]
+    with open_url(input_file_url, stream=True) as url_data:
+        for entity in ensure_dict(load_entities(file_=url_data, mimetype=mimetype)):
```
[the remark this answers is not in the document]
krebsni: For the local file used for testing url_data.headers only contains 'Content-Length'. Another option would be to try 
```suggestion
        try:
            mimetype = url_data.headers['Content-Type']
        except:
            mimetype = mimetypes.MimeTypes().guess_type(input_file_url)[0]
```
or use python-magic.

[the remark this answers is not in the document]
buehlefs: For local files the `FileAdapter` in `qhana_plugin_runner/util/request_helpers.py` can be extended to directly guess the mimetype of the file/path. This would not replace the proposed try-except, but the except clause should only listen for `KeyError`!

plugins/entity_filter.py
```diff
@@ -334,13 +401,18 @@ def entity_filter_task(self, db_id: int) -> str:
...
     if not attributes: # empty string or None
         attributes = []
-        
+    
+    # number of rows to be sampled
+    n_sampled_rows : int = 0
     if n_rows is None and not id_list:
-        raise ValueError("Row number argument not provided!")
+        # only filter columns
+        id_list = {}
```
buehlefs, line 409: Empty `{}` always result in a dict. An empty set must use `set()` constructor. Also this would be better if id_list is only used for the id list and another name for the set (`id_set`) to clearly differentiate the types. The type of id_list is also wrong (is `Optional[str]` but should be `Optional[List[str]]`).
The whole conversion can also be streamlined:

```python
id_set: Set[str] = set(params.get("id_list", []))
```

plugins/entity_filter.py
```diff
@@ -355,46 +427,28 @@ def entity_filter_task(self, db_id: int) -> str:
...
+                                          n_sampled_rows=n_sampled_rows, row_sampling=row_sampling)
+            if len(output_entities) != n_rows:
+                raise ValueError("Number of rows requested is greater than number of rows in input file!")
+                # TODO: maybe log an error instead of exception
 
-    output_entities = output_entities_id_list + output_entities_random_rows
-    if len(output_entities) != n_rows:
-        raise ValueError("Number of rows requested is greater than number of rows in input file!")
+    else:
+        # TODO take all rows
```
buehlefs, line 450: It should be easy to switch off sampling and use INFINITY for n_rows to implement this case with minimal code changes.
Alternatively make sure that the second part of the pipeline (filtering columns) can work on streams/iterables and just use the entities iterable as is without row filtering.

plugins/entity_filter.py
```diff
@@ -382,105 +434,91 @@ def entity_filter_task(self, db_id: int) -> str:
...
-    elif attributes_setting == AttributeType.BLOCKLIST.value:
-        if "ID" in attributes:
-            attributes = attributes.remove("ID")
-    else:
-        raise ValueError("Invalid argument for Attribute Setting!")
+        # Filter rows
+        if n_rows != INFINITY:
+            r_filtered_entities = filter_rows(
+                input_entities=input_entities, id_set=id_set, 
+                n_sampled_rows=n_sampled_rows, row_sampling=row_sampling)
```
buehlefs, line 497: Hast du den Black autoformatter aktiv?

qhana_plugin_runner/util/request_helpers.py
```diff
@@ -63,7 +64,11 @@ def send(self, request: PreparedRequest, stream: bool, **kwargs) -> Response:
                 resp.headers["Content-Length"] = str(file_path.stat().st_size)
                 resp.status_code = HTTPStatus.OK
 
-                # TODO set mimetype in response?
+                # set mimetype in response if guessable
+                mimetype = mimetypes.MimeTypes().guess_type(url=file_path)[0]
+                if mimetype:
+                    resp.headers["Content-Type"] = mimetype
```
buehlefs, line 70:
```suggestion
                mimetypes = mimetypes.MimeTypes().guess_type(url=file_path)
                if mimetypes:
                    resp.headers["Content-Type"] = mimetypes[0]
```
Should be safer as the index access happens after the not none/empty check.

buehlefs (commented):
Looks good, but please run the formatter.

buehlefs (approved):
Looks good to me

PR #9, opened 2022-02-22
Update dockerfile

buehlefs:
Exclude plugins from default docker build.
Allow plugins to be loaded from git into docker container.
Migrate DB on container startup.
Add container startup tasks in tasks.py.

TODO:

- [ ] Documentation for dockerfile
- [ ] Automatic docker build using GH Actions

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,52 @@
-FROM ubuntu:focal
+FROM python:3.9-slim
+
+LABEL org.opencontainers.image.source="hhttps://github.com/UST-QuAntiL/qhana-plugin-runner"
+
+# install git and remove caches again in same layer
+ARG DEBIAN_FRONTEND=noninteractive
+RUN apt-get update && apt-get install -y --no-install-recommends git && apt-get clean && rm -rf /var/lib/apt/lists/*
+
 WORKDIR /app
 
-RUN apt-get -y update && apt-get install -y python3 python3-distutils python3-apt wget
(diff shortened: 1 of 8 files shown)


Dockerfile
```diff
@@ -1,15 +1,52 @@
-FROM ubuntu:focal
+FROM python:3.9-slim
+
+LABEL org.opencontainers.image.source="hhttps://github.com/UST-QuAntiL/qhana-plugin-runner"
```
PhilWun, line 3: The url has one "h" too many.

Dockerfile
```diff
@@ -1,15 +1,52 @@
...
+# create and activate virtualenv as gunicorn user to allow installing additional dependencies later
+ARG VIRTUAL_ENV=/venv
+RUN python -m venv ${VIRTUAL_ENV}
+
+# change path to include virtualenv first (affects all following commands)
+ENV PATH="${VIRTUAL_ENV}/bin:$PATH"
+
+RUN python -m pip install --upgrade pip && python -m pip install gunicorn poetry invoke
+
+COPY --chown=gunicorn . /app
```
PhilWun, line 44: Plugins are still copied from the plugins folder into the Docker Container. Is this intentional?

buehlefs: No, plugins should be ignored as per dockerignore file.

Dockerfile
buehlefs: Should be fixed in all three repositories now.

PR #10, opened 2022-02-28
Manual classification

krebsni:
Manual classification plugin now working, bugfix for hybrid ae plugin


plugins/hybrid_ae_pkg/hybrid_ae_templates/hybrid_ae_template.html
```diff
@@ -0,0 +1,51 @@
+{% import 'forms.html' as forms %}
+
+<html>
+
+<head>
+    <link rel="stylesheet" type="text/css" href="{{url_for('static', filename='forms.css')}}" media="screen" />
+
+    <style type="text/css">
+        @import url("{{url_for('static', filename='forms.css')}}");
+    </style>
```
buehlefs, line 10: The style tag can be removed as the style sheet is already loaded with the link in line 6. Using a css import in a style tag was neccessary with the old style micro frontends as they were only html fragments.

plugins/hybrid_ae_pkg/hybrid_ae_templates/hybrid_ae_template.html
```diff
@@ -0,0 +1,51 @@
...
+</head>
+
+<body>
+    <div class="qhana-microfrontend">
+
+        <details class="qhana-help">
```
buehlefs, line 16: Some of this help text can be moved into the input model metadata (the descriptions are now displayed below the form fields and support [limited] markdown).

plugins/hybrid_ae_pkg/hybrid_ae_templates/hybrid_ae_template.html
```diff
@@ -0,0 +1,51 @@
...
+                <dt>QNN Name</dt>
+                <dd>The name of the quantum-neural-network to use as the autoencoder. <code class="example">QNN3</code>
+                </dd>
+                <dt>Training Steps</dt>
+                <dd>The number of training steps to train the autoencoder. <code class="example">100</code></dd>
+                <dl>
+
+                    <a href="{{example_values}}">Fill in example values.</a>
+
+                    <!-- TODO: Detailed description of the form options.-->
```
buehlefs, line 37: It may be possible to use the simple template if the help text is moved into the model.
We could also allow the simple template to optionally display a help text (that uses markdown).

plugins/manual_classification/manual_classification_templates/manual_classification_template.html
```diff
@@ -1,25 +1,66 @@
...
-    {% for id in id_list %}
-    <div class="qhana-form-field" style="width: 3em; height: 2em; padding: 1em;">
-        <input type="checkbox" id={{id}} name={{id}}>
-        <label class="qhana-form-label" for="{{id}}">{{id}}</label>
+<html>
+
+<head>
+    <link rel="stylesheet" type="text/css" href="{{url_for('static', filename='forms.css')}}" media="screen" />
+    <style type="text/css">
+        @import url("{{url_for('static', filename='forms.css')}}");
```
buehlefs, line 7: this line is now obsolete (see comment above)

plugins/manual_classification/manual_classification_templates/manual_classification_template.html
```diff
@@ -1,25 +1,66 @@
...
+
+        table {
+            font-family: arial, sans-serif;
```
buehlefs, line 10: do not specify fonts if possible to allow for easier font overrides in the loaded micro frontend

plugins/manual_classification/manual_classification_templates/manual_classification_template.html
```diff
@@ -1,25 +1,66 @@
...
+                    {% endfor %}
+                </tr>
+                {% endfor %}
+            </table>
+
+        </div>
+        <div class="qhana-form-buttons">
+            {{ forms.submit("validate")}}
+            {{ forms.submit("continue", action=process)}}
+            {{ forms.submit("done", action=done)}}
```
buehlefs, line 58: We really need to specify how repeatable substeps should work...

plugins/manual_classification/routes.py
```diff
@@ -48,33 +49,38 @@
...
             description="Manually annotate classes for data sets from MUSE database.",
             name=ManualClassification.instance.name,
             version=ManualClassification.instance.version,
-            type=PluginType.simple,
+            type=PluginType.complex,
             entry_point=EntryPoint(
-                href=url_for(f"{MANUAL_CLASSIFICATION_BLP.name}.LoadView"),
-                ui_href=url_for(f"{MANUAL_CLASSIFICATION_BLP.name}.MicroFrontend"),
-                data_input=[  # TODO: only file input (entities...)
+                href="./load/",
```
buehlefs, line 63: please use the `url_for` function if possible to avoid renames breaking this.

plugins/manual_classification/routes.py
```diff
@@ -154,18 +158,26 @@ def post(self, arguments):
...
+        db_task.data["step_id"] = 1
 
         # all tasks need to know about db id to load the db entry
         task: chain = pre_render_classification.s(db_id=db_task.id) | add_step.s(
-            db_id=db_task.id, step_id=step_id, href=href, ui_href=ui_href, prog_value=30
+            db_id=db_task.id,
+            step_id=step_id,
+            href=href,
+            ui_href=ui_href,
+            prog_value=db_task.data["step_id"] / (db_task.data["step_id"] + 1) * 100,
```
buehlefs, line 180: please use `()` it makes it more readable and the intent more clear `(a/b)*100` vs. `a/b*100`

plugins/manual_classification/routes.py
```diff
@@ -306,7 +334,9 @@ def post(self, arguments, db_id: str):
             TASK_LOGGER.error(msg)
             raise KeyError(msg)
 
+        TASK_LOGGER.error(">>>>>>>>>>" + str(arguments))
```
buehlefs, line 337: remove debug output or change log severity to debug

plugins/manual_classification/tasks.py
```diff
@@ -103,16 +110,17 @@ def add_class(self, db_id: int) -> str:
         TASK_LOGGER.error(msg)
         raise ValueError(msg)
 
-    entity_data = task_data.data["entity_data"]
-    for id in entity_data.keys():
+    entity_annotation = copy.deepcopy(task_data.data["entity_annotation"])
```
buehlefs, line 113: I think it is possible to mark this as changed without a deep copy but I may be wrong. Either way this is something that deserves a warning in the documentation!

buehlefs (changes requested):

plugins/hybrid_ae_pkg/hybrid_ae_templates/hybrid_ae_template.html
krebsni: Ok, I fixed it. Probably not necessary to have the example values included at all as there is the option to fill out example values.

plugins/hybrid_ae_pkg/hybrid_ae_templates/hybrid_ae_template.html
krebsni: Yes that makes more sense. Moved to simple template in the next commit.

plugins/manual_classification/manual_classification_templates/manual_classification_template.html
krebsni: Should we include the above way in the spec or is it still open for discussion?

buehlefs: For now we could include it in the simple template (gated by an if that checks if there is a done URL defined). The micro frontend can find the done URL by internal information. But for other API clients we need to specify how the done URL can be found (without scraping the micro frontend). There are some plans but nothing final, so no spec changes for now. (But you culd start an ADR to sketch some solutions)

krebsni: We could also include the above functionality in simple template in case "done" is specified

buehlefs: Yes. See the first sentence of my previous reply.

plugins/manual_classification/tasks.py
krebsni: The problem is that we use SQL Alchemy's JSON type in the key-value storage (for type of the value) that does not detect in-place mutations. Hence, the need of a deep copy. If you can think of a better solution, let me know. Else, I have noted it in the documentation.

plugins/manual_classification/manual_classification_templates/manual_classification_template.html
```diff
@@ -3,64 +3,52 @@
...
         {% call forms.render_form() %}
         <!-- class name -->
         {{ forms.render_fields(schema, values=values, errors=errors) }}
-        <div class="qhana-form" style="flex-direction: row; flex-wrap: wrap;">
+        <div class="qhana-table" style="flex-direction: row; flex-wrap: wrap; 
+        height: 750px; overflow: auto;">
             <!-- select id's -->
             <table>
-                <tr>
+                <tr">
```
buehlefs, line 17:
```suggestion
                <tr>
```

plugins/manual_classification/manual_classification_templates/manual_classification_template.html
```diff
@@ -3,64 +3,52 @@
...
-                {% for entity in entity_list %}
-                <tr>
-                    <td class="qhana-form-field" style="width: 3em; padding: 1em;">
-                        <input type="checkbox" id={{entity["ID"]}} name={{entity["ID"]}}>
-                    </td>
-                    {% for attr in attr_list%}
-                    <td>{{entity[attr] if attr in entity.keys() else ""}}</td>
+                    </tr>
+                    {% for entity in entity_list %}
+                    <tr onclick="selectRow(this)">
```
buehlefs, line 24: Just letting you know: inline javascript in events like this can be disabled by the server setting a content security policy as it may be exploited to sneak in javascript. This should be considered a bad practice (or at least a code smell nowadays). But for this research prototype this is not really an issue that needs to be fixed.

plugins/manual_classification/manual_classification_templates/manual_classification_template.html
```diff
@@ -3,64 +3,52 @@
...
+            {{ forms.submit("validate", target="microfrontend")}}
             {{ forms.submit("continue", action=process)}}
             {{ forms.submit("done", action=done)}}
         </div>
         {% endcall %}
     </div>
 
     <script src="{{url_for('static', filename='microfrontend.js')}}"></script>
+    <script type="text/javascript">
+        function selectRow(row) {
```
buehlefs, line 47: Follow up from above: If a content security policy would be set by the server then the script tag would need to inlude a nonce (that should be generated new for each request) that is whitelisted in the CSP to allow the script to run. The selectRow function would have to be bound to the html elements using the `addEventListener` methods of the elements. (again, no changes needed her, just letting you know)

plugins/manual_classification/routes.py
```diff
@@ -92,7 +91,7 @@ class MicroFrontend(MethodView):
     """Micro frontend for the manual classification plugin."""
 
     example_inputs = {
-        "inputFileUrl": "file:///<path_to_file>/entities.json",
+        "inputFileUrl": "file:///home/nico/entities2.json",
```
buehlefs, line 94: We probably should not use file:// urls as examples.

qhana_plugin_runner/db/models/tasks.py
```diff
@@ -69,7 +69,7 @@ class TaskData:
     Attributes:
         id (int): ID of corresponding :class:`ProcessingTask` entry. Use the id to fetch this information from the database.
         key (str): a key in dict
-        value (str, optional): a corresponding value in dict
+        value (Union[dict, list, str, float, int, bool, None], optional): a corresponding value in dict. Note that SQL Alchemy's JSON type is used which does not support in-place mutations. If a value needs to be changed, e.g., use `copy.deepcopy`.
```
buehlefs, line 72: We should test if an actual deepcopy is needed here before advising users to always use deepcopy. It should not matter much for performance as it will probably only be used a few times per plugin call but it just does not feel right for me.

buehlefs (approved):
Please remove the extra `"` in `manual_classification_template.html` before merging.

PR #13, opened 2022-04-04
Fix import errors of plugins

PhilWun:
This PR fixes the import errors when loading plugins from a folder other than the plugins folder. This happens e.g. when using the QHAna Docker setup.

To fix this issue, the absolute imports in the plugins, that imported files from the plugin itself, were replaced by relative imports.

### DIFF
--- a/plugins/hybrid_ae_pkg/backend/quantum/pl/QNN1.py
+++ b/plugins/hybrid_ae_pkg/backend/quantum/pl/QNN1.py
@@ -3,7 +3,7 @@ from typing import Tuple, Callable, List
 import pennylane as qml
 import pennylane.numpy as np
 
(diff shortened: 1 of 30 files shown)


plugins/hybrid_ae_pkg/backend/quantum/pl/QNN1.py
```diff
@@ -3,7 +3,7 @@
 import pennylane as qml
 import pennylane.numpy as np
 
-from plugins.hybrid_ae_pkg.backend.quantum.pl.TwoQubitGate import add_two_qubit_gate
+from .TwoQubitGate import add_two_qubit_gate
```
buehlefs, line 6: if `TwoQubitGate` is a python module, then it should use `snake_case`. This could be changed in this PR or add an issue.

buehlefs (changes requested):

buehlefs (approved):

PR #15, opened 2022-04-12
File upload plugin

PhilWun:
This PR adds a plugin that allows users to upload files from their local filesystem to the file store of the plugin runner.


buehlefs (changes requested):
I've tried testing the plugin with the docker-compose setup but the file upload fails in Vivaldi (chromium based). We may need to extend the microfrontend js script to look for file inputs in a form and handle the form data differently if they are present (e.g. set the content type to multipart...)

buehlefs (approved):

PR #19, opened 2022-05-06
Feature/key value storage 

krebsni:
- now uses custom mutable JSON 
- includes unit test


qhana_plugin_runner/db/models/mutable_json.py
```diff
@@ -0,0 +1,300 @@
+# Copyright (c) 2014, Elmer de Looff <elmer.delooff@gmail.com>
```
buehlefs, line 1: Is it possible to install this from PyPi or from the github repo as a dependency? I will need to think about how to best include this sourcecode (and its licence).

tests/db_tests/conftests.py
```diff
@@ -0,0 +1,49 @@
...
+from qhana_plugin_runner.db.models.mutable_json import (
+    NestedMutableDict,
+    NestedMutableList,
+)
+
+
+@pytest.fixture(scope="function")
+def task_data():
+
+    environ["SQLALCHEMY_DATABASE_URI"] = "sqlite:///../instance/test.db"
```
buehlefs, line 39: There is also the option for an in memory sqite DB. Because a test DB should always be in a defined state at test start either create the DB in a temporary folder (use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory as a context manager for automatic cleanup) or make sure to clean up the test database completely (e.g. delete the sqlite file)

tests/db_tests/test_db.py
```diff
@@ -0,0 +1,60 @@
...
+from qhana_plugin_runner.db.models.tasks import ProcessingTask
+from qhana_plugin_runner.db.models.mutable_json import (
+    TrackedDict,
+    TrackedList,
+)
+
+
+def test_json_strorage(task_data: ProcessingTask):
+    # Test mutable json data attribute
+    assert task_data.data == dict()
```
buehlefs, line 12: please add descriptions for when the assertion fails:
```suggestion
    assert task_data.data == dict(), "Task data.data is expected to be an empty dict at the start of the unit test."
```

tests/db_tests/test_db.py
```diff
@@ -0,0 +1,60 @@
...
+
+    # Primitive data types
```
buehlefs, line 14: Please do these in individual test functions. Rule of thumb should be 1 test function tests 1 thing.

tests/db_tests/test_db.py
```diff
@@ -0,0 +1,60 @@
+from conftests import task_data
```
buehlefs, line 1: Also add the license header here

tests/db_tests/conftests.py
```diff
@@ -0,0 +1,49 @@
...
+    app = create_app()
```
buehlefs, line 40: create app can receive a test config object (I have never tested this but this should be preferred over messing with the current environment variables)

buehlefs (changes requested):

qhana_plugin_runner/db/models/mutable_json.py
krebsni: We do need to overwrite NestedMutable in order to accept primitive types without nesting. It should be possible though to import NestedMutableDict and NestedMutableList (which is basically all that is required from the lib) and add it as a dependency. That would reduce the file by about two thirds.

qhana_plugin_runner/db/models/mutable_json.py
```diff
@@ -0,0 +1,112 @@
...
+
+
+class NestedMutable(Mutable):
+    """SQLAlchemy `mutable` extension with nested change tracking."""
+
+    @classmethod
+    def _listen_on_attribute(cls, attribute, coerce, parent_cls):
+        """Overwrite method of base class in order to accept primitive (immutable) types. 
+        
+        Code is identical to MutableBase._listen_on_attribute apart from handling ``str``, ``float`` and ``int`` (which includes ``bool`` as ``bool`` is a subclass of ``int``)."""
```
PhilWun, line 31: Check if the license of the copied / modified code is compatible with the Apache 2.0 license. If that's the case, satisfy the conditions of the license e.g. include the license notice and permission notice.

PhilWun (approved):
lgtm

buehlefs (approved):

PR #21, opened 2022-05-19
Add GitHub workflow to build Docker images for the mini MUSE database

PhilWun:
With this PR, images for the mini MUSE database can be build with a GitHub workflow. This workflow needs to be manually triggered to avoid unnecessary builds and because changes to the mini MUSE database are expected to be infrequent.

### DIFF
--- a/.github/workflows/mini_muse_image.yml
+++ b/.github/workflows/mini_muse_image.yml
@@ -1,0 +1,39 @@
+name: Build and Push Mini MUSE Image
+on: [workflow_dispatch]
+
+env:
+    REGISTRY: ghcr.io
+    IMAGE_NAME: ${{ github.repository_owner }}/mini-muse
+
+jobs:
+  build_image:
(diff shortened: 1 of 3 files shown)


.github/workflows/mini_muse_image.yml
```diff
@@ -0,0 +1,39 @@
...
+      - name: Login to GitHub Container Registry
+        uses: docker/login-action@v2
+        with:
+          registry: ${{ env.REGISTRY }}
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v3
+        with:
+          context: ./plugins/costume_loader_pkg/db_container/
```
buehlefs, line 35: We need to keep that in mind when we ver move that plugin.

buehlefs (approved):
lgtm

PR #31, opened 2022-08-01
Return representative circuit from quantum k-means

PhilWun:


plugins/quantum_k_means/tasks.py
```diff
@@ -140,4 +140,14 @@ def calculation_task(self, db_id: int) -> str:
             "application/json",
         )
 
+    with SpooledTemporaryFile(mode="w") as output:
+        output.write(representative_circuit)
+        STORE.persist_task_result(
+            db_id,
+            output,
+            "representative_circuit.qasm",
+            "representative-circuit",
```
buehlefs, line 149: We need to define a good data type for quantum circuits in the documentation.

buehlefs (approved):
lgtm

Issue #29, opened 2022-07-13
Investigate Issue with mariadb connection


buehlefs:
Investigate an Issue with mariadb connection using. The issue seems to arise from a bug either in SQLAlchemy or the database driver mysql-connector-python, sqlalchemy and possibly the mariadb version.

Known version ranges wehere the bug can occur:
mysql-connector-python > 8.0.26
sqlalchemy > 1.4.31
mariadb == 1:10.3.34-0ubuntu0.20.04.1

Known version without bugs:
mysql-connector-python == 8.0.26 (and == 8.0.25)
sqlalchemy == 1.4.31
mariadb == 1:10.3.34-0ubuntu0.20.04.1

The bug occurs when creating a new task or updating an existing task. It seems like a transaction fails to write to the database.

Possible mitigations:
- Use another database driver
- Find out which newer versions don't have the bug and upgrade to these versions
- Pin the known working versions
- Find and fix the bug

PhilWun:
Can be closed, as we no longer use mysql-connector-python.

PR #35, opened 2022-08-26
Update python dependencies

buehlefs:
* Update flask-smorest to fix incompatibility with newest flask
* Update flask and remove dependency to deprecated FLASK_ENV to detect production mode. Now use only FLASK_DEBUG to determine debug mode.
* Update dockerfile to not install into venv
* Install DB drivers as extra/optional dependencies in docker image
* Elevate invoke to a production dependency

Related to #29 as possible workaround. This PR adds a different database driver `PyMySQL` to the docker image.


PhilWun (approved):
lgtm

krebsni:
Seems to work. No problems with issue #29 on my side in tests with mariadb.

PR #40, opened 2022-09-29
Move TaskResponseSchema from the plugins to the plugin runner

PhilWun:


buehlefs:
Please check if the schema can be removed completely

PR #41, opened 2022-09-29
Refactor entity to tuple parsing to be more robust

buehlefs:
Use a class derived from namedtuple to store the actual unprocessed attribute names while still allowing for namedtuple like behaviour.

The new behaviour is more robust as attribute names that are not allowed for namedtuples can still be used without any change to the attribute name. Tuples have a `get` method that allows value access via attribute name.


PhilWun (approved):
lgtm

PR #43, opened 2022-09-30
Fix database connection error in costume loader plugin

PhilWun:
When using version 8.0.30 of the mysql-connector-python package to connect to a MariaDB database, the error "Character set 'utf8' unsupported" will occur. To prevent this, the version was pinned to 8.0.29.

### DIFF
--- a/plugins/costume_loader_pkg/backend/database.py
+++ b/plugins/costume_loader_pkg/backend/database.py
@@ -24,7 +24,7 @@ class Database(Singleton):
 
     def open_with_params(self, host: str, user: str, password: str, database: str):
         engine: Engine = create_engine(
-            "mysql+pymysql://" + user + ":" + password + "@" + host + "/" + database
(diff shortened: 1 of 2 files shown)


buehlefs:
Please check if the plugin works with the available database driver (e.g., `PyMySQL`)

plugins/costume_loader_pkg/backend/database.py
```diff
@@ -24,14 +24,7 @@ def __init__(self):
...
         engine: Engine = create_engine(
-            "mysql+mysqlconnector://"
-            + user
-            + ":"
-            + password
-            + "@"
-            + host
-            + "/"
-            + database
+            "mysql+pymysql://" + user + ":" + password + "@" + host + "/" + database
```
buehlefs, line 27: This could be an f-string

buehlefs (approved):

PR #45, opened 2022-10-07
Faster implementation of Wu-Palmer

PhilWun:
This PR replaces the Wu-Palmer implementation with a faster one and moves it from the Wu-Palmer cache generation plugin to the Wu-Palmer plugin.


plugins/wu_palmer.py
```diff
@@ -119,6 +119,16 @@ class InputParametersSchema(FrontendFormBaseSchema):
             "input_type": "textarea",
         },
     )
+    root_has_meaning_in_taxonomy = ma.fields.Boolean(
+        required=False,
+        allow_none=False,
+        metadata={
+            "label": "Root node has meaning in the taxonomy",
+            "description": "Whether the value of the root node of a taxonomy has a meaning in the taxonomy"
+            "e.g. a taxonomy about colors and the root node represents a color",
```
buehlefs, line 128: Vorschlag für eine andere Beschriftung: Consider root node part of the hierarchy.
Description wäre dann etwas wie: If the root node is part of the hierarchy, then items connected only through root are considered similar to a certain degree. Otherwise they will be considered as not similar.

Außerdem: String concatenation works in python without explicit + between literals but there must be whitespace in the strings... `...taxonomy" "e.g. a ...` should be `...taxonomy " ",e.g., a ...` (also missing commas around e.g.)

plugins/wu_palmer.py
```diff
@@ -267,10 +277,154 @@ def get_requirements(self) -> str:
...
+                self._taxonomies[tax_name]
+            )
 
+        node_paths = self._node_path_cache[tax_name]
+
+        if node_a not in node_paths:
+            raise RuntimeError(f"node {node_a} not in node paths of taxonomy {tax_name}")
+
+        if node_b not in node_paths:
+            raise RuntimeError(f"node {node_b} not in node paths of taxonomy {tax_name}")
```
buehlefs, line 356: Do not use runtime errors (instead use a more specific error or a custom error). For this case you could use KeyError (since the nodes are used as keys and are not present in the taxonomy. But key errors will already be thrown two lines later so this would only be about a better error message.

plugins/wu_palmer.py
```diff
@@ -267,10 +277,154 @@ def get_requirements(self) -> str:
...
+
+        common_ancestor_depth = 0 if self._root_has_meaning_in_taxonomy else -1
+
+        for a, b in zip(ancestors_a, ancestors_b):
+            if a != b:
+                break
+
+            common_ancestor_depth += 1
+
+        if self._root_has_meaning_in_taxonomy:
```
buehlefs, line 369: use an if at the beginning to set a variable `root_node_depth` to 0 or 1 then always subtract that depth from the denominator (x2) and the common ancestor depth (before the loop). This way it would be more clear on what is going on without more comments.

plugins/wu_palmer.py
```diff
@@ -267,10 +277,154 @@ def get_requirements(self) -> str:
...
+    wu_palmer_cache: WuPalmerCache,
+):
+    if attribute not in entity1 or attribute not in entity2:
+        return
+
+    values1 = entity1[attribute]
+    values2 = entity2[attribute]
+
+    # extract taxonomy name from refTarget
+    tax_name: str = entities_metadata[attribute]["refTarget"].split(":")[1][:-5]
```
buehlefs, line 397: Use pathlib to strip file name suffixes reliably instead of discarding the last n characters!

plugins/wu_palmer.py
```diff
@@ -267,10 +277,154 @@ def get_requirements(self) -> str:
...
+            va1, val2 = sorted((val1, val2))
+
+            if val1 is None or val2 is None:
+                sim = None
+            else:
+                sim = wu_palmer_cache.calculate_similarity(tax_name, val1, val2)
+
+            similarities[(val1, val2)] = {
+                "ID": str(val1) + "_" + str(val2),
+                "href": "",
```
buehlefs, line 418: you are allowed to omit href in entities

plugins/wu_palmer.py
```diff
@@ -267,10 +277,154 @@ def get_requirements(self) -> str:
...
+            # Wu-Palmer is commutative
+            va1, val2 = sorted((val1, val2))
+
+            if val1 is None or val2 is None:
+                sim = None
+            else:
+                sim = wu_palmer_cache.calculate_similarity(tax_name, val1, val2)
+
+            similarities[(val1, val2)] = {
+                "ID": str(val1) + "_" + str(val2),
```
buehlefs, line 417: The similarities could be stored as relations with attributes

buehlefs (changes requested):

plugins/wu_palmer.py
```diff
@@ -414,10 +420,8 @@ def add_similarities_for_entities(
                 sim = wu_palmer_cache.calculate_similarity(tax_name, val1, val2)
 
             similarities[(val1, val2)] = {
-                "ID": str(val1) + "_" + str(val2),
-                "href": "",
-                "value_1": val1,
-                "value_2": val2,
+                "source": val1,
```
buehlefs, line 423: these values were sorted. This is ok for calculating the similarity but we may want to preserve the order to correctly determine which value was from which entity later.

buehlefs (approved):

PR #48, opened 2022-10-27
Qiskit Circuit Executor

buehlefs:
Add a first circuit executor plugin using the qiskit aer statevector simulator as quantum backend.


PhilWun (approved):
lgtm

PR #50, opened 2022-11-07
NISQ Analyzer Results in QHAna

cloudsftp:


plugins/nisq_analyzer.py
```diff
@@ -86,6 +147,35 @@ def __init__(self, app: Optional[Flask]) -> None:
 
         self.url = app.config.get("NISQ_ANALYZER_UI_URL")
         self.url = environ.get("NISQ_ANALYZER_UI_URL", self.url)
+        self.url += "/#/algorithms"
```
buehlefs, line 150: Use <https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse> to work with and manipulate URLs.
Also, do not use the URL as is in the entryPoint.href but add the query parameter (that is currently added by the QHAna UI in the related PR) in the pluginsView. (see other comment.

plugins/nisq_analyzer.py
```diff
@@ -64,16 +94,47 @@ def get(self):
             version=plugin.version,
             type=PluginType.processing,
             entry_point=EntryPoint(
-                href=url_for(f"{NISQ_BLP.name}.PluginsView"),
+                href=url_for(f"{NISQ_BLP.name}.ProcessView"),
```
buehlefs, line 97: add the expected query parameter to the ui href here (using <https://docs.python.org/3/library/urllib.parse.html#module-urllib.parse>)

buehlefs (changes requested):

plugins/nisq_analyzer.py
```diff
@@ -114,7 +125,7 @@ def get(self):
 class ProcessView(MethodView):
     """Start a long running processing task."""
 
-    @NISQ_BLP.arguments(NisqAnalyzerParametersSchema(unknown=EXCLUDE), location="form")
+    @NISQ_BLP.arguments(NisqAnalyzerParametersSchema(unknown=EXCLUDE), location="json")
```
buehlefs, line 128: If you use `unknown=INCLUDE` instead, then you can use an empty schema and do not need to wrap the incoming json in an envelope. Of course, a schema that actually validates the nisq analyzer results would be even better.
(see <https://marshmallow.readthedocs.io/en/stable/quickstart.html#handling-unknown-fields>)

buehlefs (changes requested):
Just minor changes now.

buehlefs (approved):

PR #58, opened 2022-12-02
Feature/validation feedback

krebsni:


qhana_plugin_runner/templates/simple_template.html
```diff
@@ -31,6 +31,9 @@
             {% if done %}
             {{ forms.submit("done", action=done)}}
             {% endif %}
+            {% if errors == {} and valid %}
```
buehlefs, line 34:
```suggestion
            {% if not errors and valid %}
```

qhana_plugin_runner/templates/simple_template.html
```diff
@@ -31,6 +31,9 @@
...
+            <mat-icon aria-hidden="true" class="material-icon notranslate material-icons ng-star-inserted qhana-valid-icon">done</mat-icon>
```
buehlefs, line 35: This actually works? A better solution would be to inline the icon SVG. Also make sure that the valid indicator has at least some text that can be seen (and read by screenreaders) explaining what the green checkmark actually means.

buehlefs (changes requested):
I've only left comments for the first occurrence but please fix them all.

plugins/pca.py
```diff
@@ -0,0 +1,382 @@
+# Copyright 2021 QHAna plugin runner contributors.
```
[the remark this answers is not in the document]
krebsni: Not sure must have somehow slipped in when pca was refactored into own folder... I just deleted it for now.

qhana_plugin_runner/templates/simple_template.html
```diff
@@ -31,8 +31,9 @@
             {% if done %}
             {{ forms.submit("done", action=done)}}
             {% endif %}
-            {% if errors == {} and valid %}
-            <mat-icon aria-hidden="true" class="material-icon notranslate material-icons ng-star-inserted qhana-valid-icon">done</mat-icon>
+            {% if not errors and valid %}
+            <span class="qhana-input-description"> Validation successful!</span>
+            <svg class="checkmark" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 40 40"><path class="checkmark__check" fill="none" d="M14.1 27.2l7.1 7.2 16.7-16.8"/></svg>
```
buehlefs, line 36: Add source and license info in html comment

buehlefs (approved):

PR #60, opened 2022-12-11
qiskit vqc plugin

BenediktRiegel:


plugins/variational_quantum_classifier/tasks.py
```diff
@@ -0,0 +1,327 @@
...
+
+from .backend.visualization import plot_data, plot_confusion_matrix
+
+TASK_LOGGER = get_task_logger(__name__)
+
+
+def get_point(ent):
+    dimension_keys = list(ent.keys())
+    dimension_keys.remove("ID")
+    dimension_keys.remove("href")
```
buehlefs, line 53:
```suggestion
    dimension_keys = [k for k in ent.keys() if k not in ("ID", "href")]
```

plugins/variational_quantum_classifier/tasks.py
```diff
@@ -0,0 +1,327 @@
...
+    Return a generator for the entity labels, given an url to them.
+    :param entity_labels_url: url to the entity labels
+    """
+    file_ = open_url(entity_labels_url)
+    file_.encoding = "utf-8"
+    file_type = file_.headers["Content-Type"]
+    entities_generator = load_entities(file_, mimetype=file_type)
+    entities_generator = ensure_dict(entities_generator)
+    for ent in entities_generator:
+        yield {"ID": ent["ID"], "href": ent.get("href", ""), "label": ent["label"]}
```
PhilWun, line 117: The expected format of the labels should be documented somewhere, i.e. that the label needs to be in the attribute "label" and that it needs to be an integer.

plugins/variational_quantum_classifier/tasks.py
```diff
@@ -0,0 +1,327 @@
...
+    vqc_ansatz = vqc_ansatz_enum.get_ansatz(
+        n_qbits, entanglement_pattern_ansatz, reps_ansatz
+    )
+
+    # Prep optimizer
+    optimizer = optimizer_enum.get_optimizer(maxitr)
+
+    # VQC
+    vqc = qiskitVQC(backend, feature_map, vqc_ansatz, optimizer)
+    vqc.fit(train_points, train_labels)
```
PhilWun, line 211: It would be nice to display the progress of the training, because it can take a while.

PhilWun (changes requested):

plugins/variational_quantum_classifier/backend/vqc.py
```diff
@@ -0,0 +1,142 @@
...
+        )
+
+    def prep_labels(self, labels):
+        n_samples = len(labels)
+        unique_labels = list(set(labels))
+        idx_to_label = {idx: label for idx, label in enumerate(unique_labels)}
+        label_to_idx = {label: idx for idx, label in enumerate(unique_labels)}
+        n_classes = len(list(set(labels)))
+
+        # convert labels to one-hot
```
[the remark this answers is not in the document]
BenediktRiegel: vpc.predict() returns the internally computed one-hot encodings, thus I would still need to find out which one-hot encoding corresponds to which label. Additionally, I tried using labels that weren't one hot encodings (see below) and vpc.fit() threw an error for example2, but not example3, although example2 and example3 are almost identical, except that label_example3 is an already created one-hot encoding of label_example2. I did not look further into why exactly an error appeared, but I think I should just keep the code the way it is (in terms of performing my own one-hot encoding before hand). What do you think? Should I look further into why the error appears?

The dataset examples I used where:
data_example1 = np.array([
        [1, 0, 0],
        [-1, -1, 1],
        [0, 1, 0],
    ])
    label_example1 = np.array([
        "Tiger",
        "Igel",
        "Hamster",
    ])
    data_example2 = np.array([
        [-1, 0, 1],
        [1, 0, 0],
        [-1, -1, 1],
        [0, 1, 0],
    ])
    label_example2 = np.array([
        "Igel",
        "Tiger",
        "Igel",
        "Hamster",
    ])
    data_example3 = np.array([
        [-1, 0, 1],
        [1, 0, 0],
        [-1, -1, 1],
        [0, 1, 0],
    ])
    label_example3 = np.array([
        [0, 0, 1],
        [0, 1, 0],
        [0, 0, 1],
        [1, 0, 0],
    ])


The parameters for vqc where:
no. qubits = 3

feature_map = qiskit.circuit.library.ZFeatureMap,
feature_map_reps = 4
feature_map_entanglement_pattern = full

ansatz = qiskit.circuit.library.RealAmplitudes
ansatz_reps = 4
ansatz_entanglement_pattern = full

optimizer = qiskit.algorithms.optimizers.COBYLA
maxiter = 1000

backend = aer_statevector_simulator
shots = 1024

plugins/variational_quantum_classifier/tasks.py
BenediktRiegel: I couldn't realy make out a solution on how to display the training progress. There is an option for a so called "callback" function. The callback function is called every time the objective function is called. The optimizer might use the objective function multiple times during one iteration, thus I cannot trake the current iteration.
Also, when using the adam optimizer, my callback function gets called only once, after training. This is very unusual and I am not sure what to make of this.

Do you have a solution? If not, I can look further into this.

plugins/variational_quantum_classifier/tasks.py
BenediktRiegel: I changed the code and the label data type, such that labels can now be strings aswell, e.g. "dog" or "cat". I documented the data type in the docs entities.rst. I would like the labels to not be restricted to strings, but also to include integers or other data types, but this seems to be very "unclean". What do you think?

plugins/variational_quantum_classifier/backend/visualization.py
```diff
@@ -138,24 +144,25 @@ def plot_data(
...
     dimensions = ["x", "y", "z"][:dim]
     ids = (
         get_id_list(train_id_to_idx)[:train_end] + get_id_list(test_id_to_idx)[:test_end]
     )
 
-    train_labels = [str(int(el)) for el in train_labels[:train_end]]
-    test_labels = [str(int(el)) for el in test_labels[:test_end]]
-    labels = train_labels[:train_end] + test_labels[:test_end]
+    train_labels = [str(el) for el in train_labels[:train_end]]
+    test_labels = [str(el) for el in test_labels[:test_end]]
```
buehlefs, line 164: The slices in this line and the one above should also not be required

plugins/variational_quantum_classifier/tasks.py
```diff
@@ -48,16 +48,11 @@
...
-
+    dimension_keys = [k for k in ent.keys() if k not in ("ID", "href")]
     dimension_keys.sort()
     point = np.empty(
-        (
-            len(
-                dimension_keys,
-            )
+        len(
+            dimension_keys,
```
buehlefs, line 55: remove comma so that black can wrap this back to a single line

plugins/variational_quantum_classifier/backend/vqc.py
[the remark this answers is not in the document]
PhilWun: Understandable. You can keep the code the way it is and also you don't need to look further into the error.

plugins/variational_quantum_classifier/tasks.py
PhilWun: I think it would be fine to allow arbitrary data types as labels. With the label_to_int and int_to_label dicts you already have what you need for this. You probably only need to remove the string convertion `label = str(ent["label"])`.

plugins/variational_quantum_classifier/tasks.py
PhilWun: I don't know a better solution than using the callback function and since the callback function seems to be unreliable, I would say that you don't need to put more time into this.

plugins/variational_quantum_classifier/backend/visualization.py
BenediktRiegel: When creating a pandas.DataFrame all arrays must be of the same length, thus the slices are necessary.

plugins/variational_quantum_classifier/backend/optimizer.py
```diff
@@ -0,0 +1,58 @@
...
+    aqgd = "AQGD"
+    bobyqa = "BOBYQA"
+    cobyla = "COBYLA"
+    nelder_mead = "NELDER_MEAD"
+    spsa = "SPSA"
+    powell = "POWELL"
+    nft = "NFT"
+    tnc = "TNC"
+
+    def get_optimizer(self, maxitr):
```
PhilWun, line 40: please add type hints to function parameters

plugins/variational_quantum_classifier/backend/visualization.py
```diff
@@ -0,0 +1,275 @@
...
+
+def get_id_list(id_to_idx: dict) -> list:
+    ids = ["id"] * len(id_to_idx)
+    for id, idx in id_to_idx.items():
+        ids[idx] = id
+    return ids
+
+
+def add_background(
+    points, resolution, predictor, scatter, two_classes=False, label_to_int=None
```
PhilWun, line 30: type hints

plugins/variational_quantum_classifier/backend/visualization.py
```diff
@@ -0,0 +1,275 @@
...
+
+    # Set x- and y-axes correctly, in case the background still is not large enough for autozoom
+    scatter.update_xaxes(range=[x_grid.min(), x_grid.max()])
+    scatter.update_yaxes(range=[y_grid.min(), y_grid.max()])
+
+    return scatter
+
+
+def plot_data(
+    train_data,
```
PhilWun, line 137: type hints

plugins/variational_quantum_classifier/backend/visualization.py
```diff
@@ -0,0 +1,275 @@
...
+                    size=30,
+                ),
+            )
+        )
+    )
+
+    return fig
+
+
+def plot_confusion_matrix(y_true, y_pred, labels: list):
```
PhilWun, line 234: type hints

plugins/variational_quantum_classifier/backend/vqc.py
```diff
@@ -0,0 +1,142 @@
...
+        else:
+            err_str = f"Unkown ansatz {self.name}!"
+            raise ValueError(err_str)
+
+
+class QiskitVQC:
+    def __init__(
+        self,
+        quantum_instance: QuantumInstance,
+        feature_map,
```
PhilWun, line 62: type hints

plugins/variational_quantum_classifier/tasks.py
```diff
@@ -0,0 +1,343 @@
...
+from .backend.vqc import QiskitVQC
+
+from sklearn.metrics import accuracy_score
+
+from .backend.visualization import plot_data, plot_confusion_matrix
+
+TASK_LOGGER = get_task_logger(__name__)
+
+
+def get_point(ent):
```
PhilWun, line 50: type hint

PhilWun (changes requested):

PhilWun (approved):

PR #61, opened 2022-12-23
Update plugin types and data types

krebsni:
- [x] plugin types
- [x] data types

Uncertain about some data types: 

- costume loader: `entity/list` statt raw and `custom/taxonomy-graphs` statt `graphs `(output)?
- zip merger: `custom/any` statt `any`?
- visualization: `entity/numeric` statt `entity-points` (input)?
- one-hot encoding: `entity/vector` statt `entity-points` (output)?
- json/csv visualization, file upload: change from `*`?
- custom types for wu palmer, time tanh, sym max mean, transformers, aggregators, mds

### DIFF
--- a/default-launch.json
+++ b/default-launch.json
@@ -2,7 +2,7 @@
     // Use IntelliSense to learn about possible attributes.
     // Hover to view descriptions of existing attributes.
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
-    "version": "0.2.0",
+    "version": "0.3.0",
     "compounds": [
         {
             "name": "All",
--- a/default-launch.json
+++ b/default-launch.json
@@ -2,7 +2,7 @@
     // Use IntelliSense to learn about possible attributes.
     // Hover to view descriptions of existing attributes.
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
-    "version": "0.3.0",
+    "version": "0.2.0",
     "compounds": [
         {
             "name": "All",
--- a/plugins/quantum_k_means/__init__.py
+++ b/plugins/quantum_k_means/__init__.py
@@ -20,7 +20,7 @@ from qhana_plugin_runner.api.util import SecurityBlueprint
 from qhana_plugin_runner.util.plugins import plugin_identifier, QHAnaPluginBase
 
 _plugin_name = "quantum-k-means"
-__version__ = "v1.2.0"
+__version__ = "v0.2.0"
 _identifier = plugin_identifier(_plugin_name, __version__)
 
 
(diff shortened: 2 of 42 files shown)


buehlefs:
> Uncertain about some data types:
> 
> * costume loader: `entity/list` statt raw and `custom/taxonomy-graphs` statt `graphs `(output)?

This should be `graph/taxonomy` (see <https://qhana-plugin-runner.readthedocs.io/en/latest/data-formats/examples/graphs.html>)

> * zip merger: `custom/any` statt `any`?

If no specific data type is required here then use `*`

> * visualization: `entity/numeric` statt `entity-points` (input)?

Numeric should fit. We may need to use the less specific `entity/*` form for now to catch all numeric entity types until the tooling is smart enough to recognise equivalent data types.

> * one-hot encoding: `entity/vector` statt `entity-points` (output)?

It should be `entity/vector`, yes

> * json/csv visualization, file upload: change from `*`?

No, these plugins are too generic and the `*` fits. The file upload plugin should not have a data input field though...

> * custom types for wu palmer, time tanh, sym max mean, transformers, aggregators, mds

Wherever possible the `entity/*` types should be used but when yu can't decide this it is fine with me to use a custom type for now.

qhana_plugin_runner/api/plugin_schemas.py
```diff
@@ -33,8 +33,6 @@ class PluginType(Enum):
         "simple" and "complex" are deprecated and should be replaced with "processing"!
```
buehlefs, line 33: Also adjust the doc string of the enum.

buehlefs (changes requested):
Looks good aside from some version numbers having skipped a few versions (wich is just a minor cosmetic issue).
See the previous comment for what to do with the remaining open questions.
Maybe we should also start using the changelog file in the repo. We could mention the deprecation in there.

default-launch.json
```diff
@@ -2,7 +2,7 @@
     // Use IntelliSense to learn about possible attributes.
     // Hover to view descriptions of existing attributes.
     // For more information, visit: https://go.microsoft.com/fwlink/?linkid=830387
-    "version": "0.2.0",
+    "version": "0.3.0",
```
buehlefs, line 5: This version has nothing to do with the plugin runner version! The config here is for the VSCode Debugger.

plugins/quantum_k_means/__init__.py
```diff
@@ -20,7 +20,7 @@
 from qhana_plugin_runner.util.plugins import plugin_identifier, QHAnaPluginBase
 
 _plugin_name = "quantum-k-means"
-__version__ = "v1.2.0"
+__version__ = "v0.2.0"
```
buehlefs, line 23: skipping to a major version was ok, but here there was a major+minor version change, hence the comment about skipping multiple versions. But it is ok to leave it like it is now.

buehlefs (changes requested):
Please adjust the plugin runner version in the correct file (`pyprojct.toml`)

buehlefs (approved):

PR #64, opened 2023-02-02
Improve workflow plugin readme, add plugin registry to docker compose

PhilWun:


docker-compose.yml
```diff
@@ -51,9 +57,76 @@ services:
...
+      PLUGIN_REGISTRY_URL: "http://registry:8080"
+  backend:
+    image: ghcr.io/ust-quantil/qhana-backend:main
+    volumes:
+      - experiments:/app/data
+    extra_hosts:
+      - "host.docker.internal:host-gateway"
+    environment:
+      # remove the following line if you're running the plugin-runner and worker outside of docker
+      QHANA_URL_MAPPING: '{"(?<=^|https?://)(localhost|host.docker.internal):5005": "qhana-plugin-runner:8080", "(?<=^|https?://)localhost": "host.docker.internal"}'
```
buehlefs, line 69: We need a better solution for this mess in the future... (but I am not sure if there is a better solution)

docker-compose.yml
```diff
@@ -51,9 +57,76 @@ services:
       WAIT_HOSTS: redis:6379, postgres:5432, camunda:8080
       WAIT_TIMEOUT: 300
       WAIT_SLEEP_INTERVAL: 5
+      PLUGIN_REGISTRY_URL: "http://registry:8080"
+  backend:
```
buehlefs, line 61: I am not sure if we should include UI and backend here. We also may want to use profiles to only start the additional components when actually required.

buehlefs (commented):
Also extend the tasks.py file with commands to start the plugin registry containers on demand for development (the command can just wrap the corresponding docker/docker-compose calls).

docker-compose.yml
```diff
@@ -17,6 +17,7 @@ services:
       SQLALCHEMY_DATABASE_URI: "postgresql+psycopg2://user:password@postgres:5432/default_db"
       CAMUNDA_API_URL: "http://host.docker.internal:8081/engine-rest"
       PLUGIN_RUNNER_URLS: "http://host.docker.internal:5005"
+      LOCALHOST_PROXY_PORTS: ":9090"
```
buehlefs, line 20: The plugin runner also needs to access the registry (and the camunda workflow tool).

docker-compose.yml
```diff
@@ -48,12 +55,71 @@ services:
...
       PERIODIC_SCHEDULER: "True"
+      LOCALHOST_PROXY_PORTS: ":9090"
       WAIT_HOSTS: redis:6379, postgres:5432, camunda:8080
       WAIT_TIMEOUT: 300
       WAIT_SLEEP_INTERVAL: 5
+      PLUGIN_REGISTRY_URL: "http://registry:8080"
+  backend:
+    image: ghcr.io/ust-quantil/qhana-backend:main
+    profiles:
+      - with_backend
```
buehlefs, line 66: add the with_ui profile so that the ui will always have the backend

plugins/workflows/README.md
```diff
@@ -153,6 +155,19 @@ Currently, following exception error codes are supported:
...
 You can find examples in the `/bpmn` directory of the workflows plugin to get started.
 
+## Deploying Workflows
+
+Open the workflow in Camunda Modeler.
+This can be your own workflow or one the example workflows.
+- click on the rocket icon at the bottom
+- enter a deployment name
+- set the REST endpoint to the endpoint of Camunda BPMN
+  - if you're using the docker setup this would be http://localhost:8081/engine-rest
```
buehlefs, line 165: we may want to choose a different port as 8081 is an often used port

docker-compose.yml
PhilWun: These two are already configured in other env vars to be access via host.docker.internal. Or should I change them to localhost and add the ports to the proxy config?

buehlefs: For the registry using host.docker.internal should not cause any problems, but with camunda, we may end up accessing it with both URLs/passing URLs to the UI. I think it is better to be fully consistent to avoid confusion. So everything (besides the databases) should be localhost.

buehlefs (approved):

PR #67, opened 2023-02-28
Fix data format of the MDS plugin

PhilWun:
This PR fixes the format of the output data of the MDS plugin. The output is now compatible with the specification of the `entity/vector` data type and can be used with the new version of the quantum k-means plugin.

### DIFF
--- a/plugins/mds.py
+++ b/plugins/mds.py
@@ -360,9 +360,10 @@ def calculation_task(self, db_id: int) -> str:
     entity_points = []
 
     for ent_id, idx in id_to_idx.items():
-        entity_points.append(
-            {"ID": ent_id, "href": "", "point": [x for x in transformed[idx]]}
-        )
+        new_entity_point = {"ID": ent_id, "href": ""}
+        new_entity_point.update({str(i): x for i, x in enumerate(transformed[idx])})
+
+        entity_points.append(new_entity_point)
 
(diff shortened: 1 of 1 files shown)


plugins/mds.py
```diff
@@ -360,9 +360,10 @@ def calculation_task(self, db_id: int) -> str:
     entity_points = []
 
     for ent_id, idx in id_to_idx.items():
-        entity_points.append(
-            {"ID": ent_id, "href": "", "point": [x for x in transformed[idx]]}
-        )
+        new_entity_point = {"ID": ent_id, "href": ""}
+        new_entity_point.update({str(i): x for i, x in enumerate(transformed[idx])})
```
buehlefs, line 364: The entity attributes are sorted lexicographically. This means that dimensions will sort like this `1,11,12,2,3,4`. Number strings have to be padded with 0 to sort correctly. Other plugins outputting `entity/vector` data already do this.

buehlefs (changes requested):

plugins/mds.py
```diff
@@ -376,3 +377,16 @@ def calculation_task(self, db_id: int) -> str:
         )
 
     return "Result stored in file"
+
+
+def _get_dim_attributes(dim):
```
buehlefs, line 382: We may want to introduce helper methods for this into the plugin runner some time.

Issue #68, opened 2023-03-10
Update dependencies


PhilWun:
Fix possible deprecation warnings, then update dependencies.
See:
- https://github.com/UST-QuAntiL/qhana-plugin-registry/pull/5
- https://github.com/buehlefs/flask-template/compare/v0.4.0...v0.4.1

PR #69, opened 2023-03-17
SQLAlchemy 2.0 + other dependency updates

PhilWun:
This PR updates most of the non-dev dependencies and converts the datamodels to the native dataclass integration of SQLAlchemy 2.0.

Closes https://github.com/UST-QuAntiL/qhana-plugin-runner/issues/68


migrations/env.py
```diff
@@ -93,19 +105,13 @@ def process_revision_directives(context, revision, directives):
...
+    connectable = get_engine()
 
     with connectable.connect() as connection:
         context.configure(
             connection=connection,
-            target_metadata=target_metadata,
+            target_metadata=get_metadata(),
             process_revision_directives=process_revision_directives,
-            render_as_batch=True,
-            compare_type=True,
```
buehlefs, line 108: compare_type is important to catch type changes of a column. Is this true by default now?

migrations/env.py
buehlefs, line 1: there are some changes here that I need to manually verify.

buehlefs (commented):
Looks good, but I need to review the changed migration script in more detail before we can merge this.

PR #70, opened 2023-03-20
Integration tests GitHub workflow

PhilWun:
This PR adds a GitHub workflow that runs the QHAna integration tests every Saturday at 12:00 pm UTC and they can also be run manually on any branch.

### DIFF
--- a/.github/workflows/integration-tests.yml
+++ b/.github/workflows/integration-tests.yml
@@ -48,6 +48,10 @@ jobs:
           registry: ${{ env.REGISTRY }}
           username: ${{ github.actor }}
           password: ${{ secrets.GITHUB_TOKEN }}
+      - id: image_tag
+        uses: ASzc/change-string-case-action@v5
+        with:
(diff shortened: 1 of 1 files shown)


.github/workflows/integration-tests.yml
```diff
@@ -0,0 +1,79 @@
...
+          username: ${{ github.actor }}
+          password: ${{ secrets.GITHUB_TOKEN }}
+      - name: Build and push Docker image
+        uses: docker/build-push-action@v4
+        id: build_push
+        with:
+          context: .
+          platforms: linux/amd64
+          push: true
+          tags: ${{ steps.meta.outputs.tags }}
```
buehlefs, line 58: Can you add a tag or label for integration test (so that we can identify and remove them easily)? Also check if these images may pile up to create issues for us later.

buehlefs (approved):

buehlefs (approved):

PR #75, opened 2023-04-05
Feature/classical k means

BenediktRiegel:
Added the plugin `classical-k-means`, which implements a classical version of the k means.


plugins/classical_k_means/schemas.py
```diff
@@ -0,0 +1,102 @@
...
+    ):
+        self.entity_points_url = entity_points_url
+        self.num_clusters = num_clusters
+        self.maxiter = maxiter
+        self.relative_residual = relative_residual
+        self.visualize = visualize
+
+    def __str__(self):
+        variables = self.__dict__.copy()
+        variables["ibmq_token"] = ""
```
PhilWun, line 47: This line doesn't seem to be needed.

PhilWun (approved):
lgtm

PR #84, opened 2023-06-06
Add signal and notify registry for new virtual plugins

buehlefs:
- [x] Add signals
- [x] Update registry client to support other http methods than get
- [ ] Add documentation for virtual plugins


PhilWun (approved):
lgtm

PR #90, opened 2023-08-01
Feature/rest api plugin

buehlefs:
A plugin to quickly integrate REST APIs as plugins in QHAna with minimal coding required.

- [x] Management API Skeleton
- [x] Virtual Plugin API Skeleton
- [x] Background tasks for virtual plugin
- [x] First successful request
- [x] First successful request + response handling
- [x] add first draft of response handling strategies (and default response strategy that fails based on response status code)
- [x] Clean up debug outputs

Cleanup and nice-to-have features will follow later (maybe over multiple PRs)

### DIFF
--- a/plugins/rest_connector/api/schemas.py
+++ b/plugins/rest_connector/api/schemas.py
@@ -28,7 +28,7 @@ from qhana_plugin_runner.api.util import FileUrl, FrontendFormBaseSchema, MaBase
 
 class ConnectorKey(Enum):
     # set data commands
-    NAME = "name" # TODO support tags
+    NAME = "name"  # TODO support tags
     DESCRIPTION = "description"
     BASE_URL = "base-url"
     OPENAPI_SPEC = "openapi-spec"
--- a/plugins/rest_connector/api/schemas.py
+++ b/plugins/rest_connector/api/schemas.py
@@ -18,7 +18,7 @@ from enum import Enum
 from typing import Dict, Iterable, Literal
 
 import marshmallow as ma
-from marshmallow import INCLUDE
+from marshmallow import INCLUDE, post_load
 from marshmallow.validate import OneOf
 from typing_extensions import Required, TypedDict
 
@@ -40,6 +40,7 @@ class ConnectorKey(Enum):
     BASE_URL = "base-url"
     OPENAPI_SPEC = "openapi-spec"
     ENDPOINT_URL = "endpoint-url"
+    ENDPOINT_METHOD = "endpoint-method"
     VARIABLES = "variables"
     REQUEST_HEADERS = "request-headers"
     REQUEST_BODY = "request-body"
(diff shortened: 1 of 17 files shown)


plugins/rest_connector/api/tasks.py
```diff
@@ -79,9 +82,27 @@ def perform_request(self, connector_id: str, db_id: int) -> None:
         timeout=20,  # TODO allow different timeouts?
     )
 
-    # FIXME response handling...
+    template_context: Dict[str, Any] = request_variables | {
+        "request": response.request,
+        "response": response.json(),
```
buehlefs, line 101: TODO: Do not use `json()` if not requested by the template. (this requires a proxy object or just passing the request).

plugins/rest_connector/api/tasks.py
```diff
@@ -79,9 +82,27 @@ def perform_request(self, connector_id: str, db_id: int) -> None:
...
+    }
+
+    for response_map in response_mapping:
+        with SpooledTemporaryFile(mode="w") as output:
+            content = render_template_sandboxed(response_map["data"], template_context)
```
buehlefs, line 106: TODO: files that have the "dereference urls" option set will have a url in the content that we need to dereference. (we could use the URL File store for this.)

plugins/rest_connector/api/tasks.py
```diff
@@ -70,12 +70,11 @@ def perform_request(self, connector_id: str, db_id: int) -> str:
     print(body)
 
     parsed_headers = parse_headers(BytesIO(headers.encode()))
+    endpoint_url = render_template_sandboxed(connector["endpoint_url"], request_variables)
```
buehlefs, line 82: TODO: We need to be able to match this to an openapi endpoint which may not be easy/possible with jinja templates...

plugins/rest_connector/api/tasks.py
```diff
@@ -70,12 +70,20 @@ def perform_request(self, connector_id: str, db_id: int) -> str:
...
+        else:
+            headers_dict[k] = v
+
     endpoint_url = render_template_sandboxed(connector["endpoint_url"], request_variables)
 
     response = request(
         method=connector["endpoint_method"],
         url=urljoin(connector["base_url"], endpoint_url),
-        headers={k: v for k, v in parsed_headers.items()},  # FIXME duplicate headers...
+        headers={k: v for k, v in parsed_headers.items()},
```
buehlefs, line 89: This needs to actually use the new headers dict in order to work...

plugins/rest_connector/api/schemas.py
```diff
@@ -107,8 +107,8 @@ class ConnectorVariable(TypedDict, total=False):
 
 
 class RequestFileDescriptorSchema(MaBaseSchema):
-    source = ma.fields.String(required=True)
-    name = ma.fields.String(required=False, allow_none=True)
+    source = ma.fields.URL(required=True)
```
buehlefs, line 110: Maybe use the FileURL field that supports more protocols

plugins/rest_connector/api/tasks.py
```diff
@@ -113,3 +120,24 @@ def perform_request(self, connector_id: str, db_id: int) -> str:
...
+
+def _download_files(
+    request_file_descriptors: List[Dict],
+) -> Dict[str, Tuple[str, BytesIO, str]]:
+    # TODO: cache downloaded files for subsequent requests?
+    request_files = {}
+
+    for i, req_file in enumerate(request_file_descriptors):
+        file_url = req_file["source"]
+        response = requests.get(file_url)
```
buehlefs, line 133: we may be able to use the response directly as a file like object if we open the request in streaming mode, but this will be more robust

PhilWun (approved):
lgtm

PR #91, opened 2023-08-01
Ba simulators

svimdvl:


buehlefs (commented):
We will address the comments on our end.

Maybe only generate a statevector result if specifically requested (e.g. via execution options). Only when the statevector requires an extra simulation step. This can be extrapolated for other result types.

buehlefs (approved):

PR #94, opened 2023-08-07
Feature/extend qknn

BenediktRiegel:
Extended the plugins QkNN and Quantum-Parzen-Window to allow duplicates in the input dataset.


PhilWun (approved):
lgtm

PR #95, opened 2023-08-17
Feature: Qiskit Executor Plugin

infacc:


infacc:
2 notes for now:
- No usage of [Qiskit Runtime sessions](https://qiskit.org/ecosystem/ibm-runtime/stubs/qiskit_ibm_runtime.Session.html) since they, as of now, only seem to support executing [Primitives](https://qiskit.org/documentation/apidoc/primitives.html). Using primitives instead of the [execute function](https://qiskit.org/documentation/stable/0.24/apidoc/execute.html) would reduce the information contained in the execution result ([QuasiDistribution](https://qiskit.org/documentation/stubs/qiskit.result.QuasiDistribution.html#qiskit.result.QuasiDistribution) instead of counts and less metadata).
- Non-blocking circuit execution requires 2 tasks: 1. starting the execution, 2. "waiting" for the result and persisting it. For this the [IBMQJob](https://qiskit.org/ecosystem/ibmq-provider/stubs/qiskit.providers.ibmq.job.IBMQJob.html) from the first task needs to be shared with the second task. Currently, the IBMQJob object is pickled and transmitted via `db_task.data`. Is pickling the IBMQJob object here the right choice?

buehlefs:
* ok, so we probably need another circuit executor type that is specific to these qiskit primitives to be able to use sessions with qiskit... That can come later
* pickling should be ok but is not ideal. In the best case it is possible to recreate the job just from the job id (and maybe the backend and api token) and have it fetch all the required info from the qiskit api. You can try if this approach works, but do not spend too much time on that.

infacc:
> * ok, so we probably need another circuit executor type that is specific to these qiskit primitives to be able to use sessions with qiskit... That can come later
> 
> * pickling should be ok but is not ideal. In the best case it is possible to recreate the job just from the job id (and maybe the backend and api token) and have it fetch all the required info from the qiskit api. You can try if this approach works, but do not spend too much time on that.

The IBMQJob is now [recreated from the IBMQ token and the job id](https://qiskit.org/ecosystem/ibm-runtime/how_to/retrieve_results.html#retrieve-job-results-at-a-later-time).

plugins/qiskit_executor/schemas.py
```diff
@@ -115,39 +109,25 @@ class InputParametersSchema(FrontendFormBaseSchema):
...
-        allow_none=False,
-        metadata={
-            "label": "Backend",
-            "description": "The quantum computer or simulator that will be used.",
-            "input_type": "select",
-        },
-    )
     ibmqToken = ma.fields.String(
-        required=False,
+        required=True,
```
buehlefs, line 113: The authentication parameters must be optional in the first step that starts the plugin execution.
See <https://qhana-plugin-runner.readthedocs.io/en/latest/plugin-types/circuit-executor.html#inputs> (`authentication details`).
Yes, this means that there are potentially two substeps, one for again asking for credentials and another to ask for the backend (if not already provided).

The authentication token (and the backend) can also already be part of the execution options. The execution options exist to make it easier to automate the circuit execution (as the circuits are often executed as part of an algorithm).

plugins/qiskit_executor/schemas.py
```diff
@@ -12,36 +12,44 @@
...
+class CircuitSelectionInputParameters:
     def __init__(
         self,
         circuit: str,
         executionOptions: str,
         shots: int,
-        backend: QiskitBackends,
         ibmqToken: str,
-        customBackend: str,
+        backend: str,
```
buehlefs, line 33: this should be `Optional[str]` if the schema has allow_none.

plugins/qiskit_executor/tasks.py
```diff
@@ -191,21 +114,113 @@ def execution_task(self, db_id: int) -> str:
...
+        raise ValueError(msg)
+    backend.shots = circuit_params.shots
+
+    circuit = QuantumCircuit.from_qasm_str(circuit_qasm)
+
+    job: IBMQJob = execute(circuit, backend, shots=execution_options["shots"])
 
-    metadata, counts, state_vector = execute_circuit(
-        circuit_qasm, backend, execution_options
+    db_task.data = dumps(
```
buehlefs, line 137: `db_task.data` should directly support storing json compatible objects like dicts without converting them to strings.

plugins/qiskit_executor/tasks.py
```diff
@@ -191,21 +114,113 @@ def execution_task(self, db_id: int) -> str:
...
+    params: CircuitSelectionInputParameters = CircuitSelectionParameterSchema().loads(
+        data["parameters"]
     )
+    execution_options = data["execution_options"]
+
+    service = QiskitRuntimeService(token=params.ibmqToken, channel="ibm_quantum")
+    job = service.job(job_id)
+
+    if not job.in_final_state():
+        raise JobNotFinished("Job not finished yet!")
```
buehlefs, line 183: Add comment that task will autoretry on this exception

buehlefs (changes requested):

plugins/qiskit_executor/schemas.py
infacc: The authentication is now optional in the first step. 
Currently, there is a maximum of 2 steps. If the user doesn't authenticate or doesn't select a backend, a second step is required. Should I implement 3 possible steps as you described?
> Yes, this means that there are potentially two substeps, one for again asking for credentials and another to ask for the backend (if not already provided).

Loading a token from the execution options isn't implemented yet. I'll add it in this PR.

buehlefs: Yes, it should be three steps ([] marks optional inputs):
1. Circuit + Options + [Credentials + Backend]
2. Credentials + [Backend] (skipped if credentials are available from first step)
3. Backend (skipped if backend is available from first step)

Everything that can be specified in the first step (apart from the circuit) must also be possible to specify with the execution options only! This also includes the backend.

You can completely ignore username + password authentication. If there is a possibility to only use a user revokable API token, we will only use that token in the plugins (handling passwords is easier to mess up).

plugins/qiskit_executor/schemas.py
```diff
@@ -12,25 +12,27 @@
...
+class CircuitParameters:
     def __init__(
         self,
         circuit: str,
-        executionOptions: str,
-        shots: int,
-        ibmqToken: str,
-        backend: str,
+        executionOptions: Optional[str],
+        shots: Optional[int],
```
buehlefs, line 33: If they allow none, then you should use `if params.shots` instead of `if params.shots != ""` to test for empty/mising values. (This will currently not produce any bugs in `routes.py`, but may lead to some surprises if the file gets refactored in the future.) VSCode and pycharm should both have a feature to quickly find references/usages of these parameters

plugins/qiskit_executor/schemas.py
```diff
@@ -12,25 +12,27 @@
...
 )
-from celery.utils.log import get_task_logger
+from qhana_plugin_runner.util.logging import redact_log_data
 
 
 TASK_LOGGER = get_task_logger(__name__)
 
 
-class CircuitSelectionInputParameters:
+class CircuitParameters:
```
buehlefs, line 28: This is a prime use case of dataclasses. (though keep in mind that dataclasses will have `__str__` and `__repr__` methods by default!)
Also true for other simple classes in this file

buehlefs (commented):
Looks good now, just a bit of polishing, and we can merge this (most of the comments should have no effect on this PR)

qhana_plugin_runner/util/plugins.py
```diff
@@ -270,3 +273,21 @@ def register_plugins(app: Flask):
                 plugin_blueprint,
                 url_prefix=f"{url_prefix}/plugins/{plugin.identifier}/",
             )
+
+
+def get_execution_options(url: str) -> dict:
```
buehlefs, line 278: I would prefer this to be closer to the entity marshalling utilities (see added imports; possibly in a new module related to execution metadata)

buehlefs (commented):
Looks good, but please move the library code (see comment)

plugins/qiskit_executor/routes.py
```diff
@@ -273,47 +332,168 @@ def render(self, db_id, data: Mapping, errors: dict, valid: bool):
...
+        )
+        params.ibmqToken = arguments.ibmqToken
+        if params.backend == "" and arguments.backend != "":
+            params.backend = arguments.backend
+        db_task.data["parameters"] = CircuitParameterSchema().dumps(params)
+
+        db_task.clear_previous_step()
+        db_task.save(commit=True)
+
+        if params.backend == "":
```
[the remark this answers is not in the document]
infacc: To add steps, the processing task, needs the corresponding `href` and `ui_href`. Unfortunately, the `url_for` function needs the [SERVER_NAME](https://flask.palletsprojects.com/en/3.0.x/config/#SERVER_NAME) outside of a request context. I'm not sure what's the best way to solve this. Is there a simple way to keep the request context for the processing celery task?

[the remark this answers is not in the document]
buehlefs: In the past we have solved this by generating the URLs before with the context still available and storing them in the task data.

qhana_plugin_runner/util/plugins.py
infacc: The function is now in [`qhana_plugin_runner.plugin_utils.metadata_utils`](https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/ebf6fc369499720dcfb3f0a715d72904673e410a/qhana_plugin_runner/plugin_utils/metadata_utils.py#L21-L36).

plugins/qiskit_executor/backend/qiskit_backends.py
```diff
@@ -12,36 +12,60 @@
...
+    backends = get_backends(ibmq_token)
+    if backends is None:
+        return None
+    return [backend.name() for backend in backends]
+
+
+def get_qiskit_backend(
+    backend: str, ibmq_token: str
+) -> Optional[IBMQBackend | IBMQSimulator]:
+    """Get the backend with the given name from the IBMQ provider. If no provider is found (e.g. invalid token), return None."""
```
buehlefs, line 68: `get_provider` can now return `None` which can cause problems in line 71 (in the `try` block)

plugins/qiskit_executor/routes.py
```diff
@@ -213,48 +211,49 @@ def post(self, arguments: CircuitParameters):
...
+            QiskitExecutor.instance.identifier,
+            f"backend_selection_href_{db_task.id}",
+            backend_href,
+        )
+        PluginState.set_value(
+            QiskitExecutor.instance.identifier,
+            f"backend_selection_ui_href_{db_task.id}",
+            backend_ui_href,
+            commit=True,
+        )
```
buehlefs, line 255: Instead of using the plugin state table (which is intended for global, not task specific state), you can just use the data attribute of the processing task to store small key value pairs (which you already do for parameters and options).

As a side note: You dump the parameters into a string, but the `data` field can hold arbitrary json serializable data, so there is no need to serialize json data into a string there. Only if you use the `parameters` attribute of the processing task directly (on the same level as the data attribute) do you need to use a string as value.

plugins/qiskit_executor/routes.py
```diff
@@ -301,11 +300,16 @@ def render(self, db_id, data: Mapping, errors: dict, valid: bool):
             msg = f"Could not load task data with id {db_id} to read parameters!"
             raise KeyError(msg)
 
-        params = CircuitParameterSchema().loads(db_task.data["parameters"])
+        params: CircuitParameters = CircuitParameterSchema().loads(
```
buehlefs, line 303: Use `load` if paramaters are no longer serialized to a string (see side note in comment before)

plugins/qiskit_executor/routes.py
```diff
@@ -433,13 +405,12 @@ def render(self, db_id, data: Mapping, errors: dict, valid: bool):
...
-        params: CircuitParameters = CircuitParameterSchema().loads(
-            db_task.data["parameters"]
-        )
         backend_parameter_schema = BackendParameterSchema()
-        backend_parameter_schema.fields["backend"].metadata["datalist"] = [
-            backend.name() for backend in get_backends(params.ibmqToken)
-        ]
+        backend_parameter_schema.fields["backend"].metadata[
+            "datalist"
+        ] = PluginState.get_value(
```
buehlefs, line 411: I have missed that in previous reviews. Does this update the global (i.e. class level) definition of the field? if not, then this can stay, if yes, then we should pass this additional datalist info seperate from the schema to the template.

plugins/qiskit_executor/routes.py
```diff
@@ -354,39 +358,7 @@ def post(self, arguments, db_id: int):
...
-            # start the execution task
-            task: chain = (
-                start_execution.s(db_id=db_task.id)
-                | result_watcher.si(db_id=db_task.id)
-                | save_task_result.s(db_id=db_task.id)
-            )
-        # save errors to db
-        task.link_error(save_task_error.s(db_id=db_task.id))
-        task.apply_async()
+        start_execution.s(db_id=db_task.id).apply_async()
```
buehlefs, line 361: It is probably a good idea to handle the case where the token is loaded from the env variable (see line 314) before this line in the `post` function so that the task will always see a correct token (or no token) and not `****`

plugins/qiskit_executor/tasks.py
```diff
@@ -58,7 +57,73 @@ def start_execution(self, db_id: int) -> str:
...
+            msg += " (invalid backend)"
+        self.replace(
+            add_step.s(
+                task_log=msg,
+                db_id=db_task.id,
+                step_id="backend-selection",
+                href=href,
+                ui_href=ui_href,
+                prog_value=1,
+                prog_target=2,
```
buehlefs, line 123: Read previous progress and adapt numbers as required for better progress feedback for the user.

qhana_plugin_runner/plugin_utils/metadata_utils.py
buehlefs, line 1: the `metadata` in the name does not really fit as the execution options are not really meta but first class data... Something like `execution_[data_]utils` would fit better.

buehlefs (changes requested):

plugins/qiskit_executor/routes.py
infacc: This uses an instance of the `BackendParameterSchema`. It shouldn't affect the global definition.

https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/ebf6fc369499720dcfb3f0a715d72904673e410a/plugins/qiskit_executor/routes.py#L408-L409

plugins/qiskit_executor/tasks.py
infacc: The plugin has 3 possible steps:
1. circuit selection
2. authentication (if not done in step 1)
3. backend selection (if not done in step 1 or 2)

I updated the progress indication such that the `progress_value` shows the number of successfully executed steps (0-3) and the `progress_target` shows the minimum number of steps that have to be done (1-3).
When the user enters neither `ibmqToken` not `backend`, the progress indicator shows `1/2`, since we don't know if step 3 is required or if the user enters the `backend` in the authentication step.
Failed steps (invalid token/backend) do not increase `progress_value` or `progress_target`.

plugins/qiskit_executor/routes.py
buehlefs: This is only true if the instance level field is a deep copy of the class level backend field. If not you are just accessing and mutating the same field object through the instance.

plugins/qiskit_executor/tasks.py
buehlefs: This one is in the if for the missing backend so it should already be step 2 or 3 depending on wether the ibm token was provided in the first step or the second one. You could also always say that this is step 3 (even if step 2 was skipped), then progress for this step would just be 2/3.

infacc: Now, the authentication step shows `1/3 steps` (only circuit selection is done) and the backend selection shows `2/3` steps done (circuit selection and authentication done).

plugins/qiskit_executor/routes.py
infacc: You were right, the global definition was updated. Now, the `values` are used to pass the backend names to the template.

https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/545f5296b581a46b70df6881c1889b2153bb5025/plugins/qiskit_executor/routes.py#L394-L410

infacc: ## Update

The current implementation for `input_type == 'text_with_datalist'` checks if the parameter schema has an attribute `datalists` of type `mapping` and then creates the datalist html tag.

https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/75d2ac40d29a44a1c91ca3964cb05e680db1ac4a/qhana_plugin_runner/templates/forms.html#L33-L43

https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/75d2ac40d29a44a1c91ca3964cb05e680db1ac4a/plugins/qiskit_executor/routes.py#L394-L396

buehlefs: is it possible to check for both, the global datalists property and a datalist in the field metadata (with priority to the global as override)? You could introduce a new helper function to keep the template looking nice (see <https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/main/qhana_plugin_runner/util/jinja_helpers.py>)

infacc: Yes. Now, it is checked if the field has a `datalist` in its metadata. If that's unsuccessful, it is checked whether the schema has a `datalists` dict with a corresponding key.

plugins/qiskit_executor/routes.py
```diff
@@ -194,6 +193,9 @@ def post(self, arguments: CircuitParameters):
 
         if not arguments.shots:
             arguments.shots = options.get("shots", 1024)
+        if arguments.ibmqToken == "****" and "IBMQ_TOKEN" in os.environ:
```
buehlefs, line 196: Is it better to set the token to "" (empty string) explicitly if no token can be loaded from the environment?

plugins/qiskit_executor/routes.py
```diff
@@ -232,27 +234,16 @@ def post(self, arguments: CircuitParameters):
...
-        )
+
+        db_task.data["urls"] = {
+            "authentication": auth_href,
+            "authentication_ui": auth_ui_href,
+            "backend_selection": backend_href,
+            "backend_selection_ui": backend_ui_href,
+        }
+
+        db_task.save(commit=True)
```
buehlefs, line 245: You can probably remove the save in line 214

plugins/qiskit_executor/routes.py
```diff
@@ -405,20 +391,17 @@ def render(self, db_id, data: Mapping, errors: dict, valid: bool):
...
 
-        backend_parameter_schema = BackendParameterSchema()
-        backend_parameter_schema.fields["backend"].metadata[
-            "datalist"
-        ] = PluginState.get_value(
-            QiskitExecutor.instance.identifier, f"backend_list_{db_task.id}", default=[]
-        )
+        # provide backend names to template
+        parameter_schema = BackendParameterSchema()
+        parameter_schema.datalists = {"backend-datalist": db_task.data["backend_names"]}
```
buehlefs, line 396: The key should be exactly the same as for the field (but you can choose either python's snake_case or json's camelCase style, as both are used interchangeable in the schemas). In short, there should be no need for a "-datalist" suffix here.

qhana_plugin_runner/util/jinja_helpers.py
```diff
@@ -91,8 +93,19 @@ def space_delimited_list(
     return " ".join(str(i) for i in items)
 
 
+def get_datalist_from_schema(schema: FrontendFormBaseSchema, key: str) -> Optional[list]:
+    """Get datalist for a field from schema or schema.datalists dict. If not found, return None."""
+    datalist = schema.fields[key].metadata.get("datalist", None)
+    if datalist:
+        return datalist
+    if hasattr(schema, "datalists") and isinstance(schema.datalists, dict):
+        return schema.datalists.get(f"{key}-datalist", None)
```
buehlefs, line 102:
```suggestion
        return schema.datalists.get(key, None)
```

buehlefs (commented):
This should be the final round of (very small) changes.

plugins/qiskit_executor/routes.py
infacc: If no token is loaded from the environment variable, the value of `arguments.ibmqToken` is already `""` (confirmed with debugger). The lines below this code snippet then explicitly set `arguments.ibmqToken` to `None` if no token is provided in the execution options. I guess we can keep this part as it is.

plugins/qiskit_executor/routes.py
infacc: Unfortunately, the `db_task.id` is needed for the URL generation (lines 216-236), which then need to be added again to the task. I haven't figured out a way around the 2 `db_task.save(commit=True)` calls.

buehlefs: To get a database ID without comitting use `DB.session.flush()`. The objects must be added to the session first (either with `db_task.save()` or `DB.session.add(db_task)`).

plugins/qiskit_executor/routes.py
buehlefs: That is not a given. Someone might type in `****` or the micro-frontend code might be changed later invalidating this assumption. While both examples are a bit contrived, the point is that relying on such complex interactions over two (or more) steps to stay like one would expect can be a bit of a gamble. In this case it would probably just lead to an unusual error message down the line so it can stay as is, but usually it is better to be explicit with error handling in every outside facing API endpoint.

plugins/qiskit_executor/routes.py
infacc: Would it make sense to replace `db_task.save(commit=True)` by `db_task.save(); DB.session.flush()`?

buehlefs: Yes. A commit commits the whole transaction. If an error occurs while trying to generate the URLs you do want to roll back all actions in the DB but if you have comitted before then this rollback does not happen automatically. In some cases comitting cannot be avoided (e.g. when a different process needs access to the DB record) but in this case we only need to commit at the end.

buehlefs (approved):

PR #97, opened 2023-09-05
Feature/classical preprocessing

BenediktRiegel:
This branch added the pandas-preprocessing plugin, allowing csv-files to be processed by pandas native methods.


stable_plugins/classical_ml/data_preparation/pandas_preprocessing/backend/pandas_preprocessing.py
```diff
@@ -0,0 +1,309 @@
...
+    :param ascending: str
+    :return: DataFrame
+    """
+    df.sort_values(by=by, ascending=ascending, inplace=True)
+    return df
+
+
+def strip_characters(
+    df: DataFrame,
+    characters: List[str] = "",
```
buehlefs, line 187:
```suggestion
    characters: Sequence[str] = "",
```

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/backend/pandas_preprocessing.py
```diff
@@ -0,0 +1,309 @@
...
+
+    new_columns = new_columns.split(",")
+    for n_c in new_columns:
+        if n_c in df.keys():
+            raise ValueError(
+                f"Columns {new_columns} already exist in the given dataframe."
+            )
+
+    df[new_columns] = df[column].str.split(by, expand=True)
+    if remove_column:
```
buehlefs, line 251: maybe check if old column name is also used in new_columns (do not remove a new column...)

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/backend/pandas_preprocessing.py
```diff
@@ -0,0 +1,309 @@
...
+    :param subset: str containing the columns separated by commas
+    :return: DataFrame
+    """
+    subset = process_subset(subset)
+    if subset is None:
+        subset = df.keys()
+
+    for k in subset:
+        if k in df:
+            if {type(el) for el in df[k]} == {str}:
```
buehlefs, line 274:
```suggestion
            if all(isinstance(el, str) for el in df[k]):
```

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/backend/pandas_preprocessing.py
```diff
@@ -0,0 +1,309 @@
...
+    if case == "upper":
+        function = lambda s: s.upper()
+    elif case == "lower":
+        function = lambda s: s.lower()
+    elif case == "title":
+        function = lambda s: s.title()
+
+    for k in subset:
+        if k in df:
+            if {type(el) for el in df[k]} == {str}:
```
buehlefs, line 306: see other comments

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/routes.py
```diff
@@ -0,0 +1,361 @@
...
+
+    @PDPreprocessing_BLP.response(HTTPStatus.OK, PluginMetadataSchema)
+    @PDPreprocessing_BLP.require_jwt("jwt", optional=True)
+    def get(self):
+        """pandas-preprocessing endpoint returning the plugin metadata."""
+
+        return PluginMetadata(
+            title="Pandas Preprocessing",
+            description=PDPreprocessing.instance.description,
+            name=PDPreprocessing.instance.identifier,
```
buehlefs, line 66:
```suggestion
            name=PDPreprocessing.instance.name,
```
See UST-QuAntiL/qhana-plugin-registry#27

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/routes.py
```diff
@@ -0,0 +1,361 @@
...
+            version=PDPreprocessing.instance.version,
+            type=PluginType.processing,
+            entry_point=EntryPoint(
+                href=url_for(f"{PDPreprocessing_BLP.name}.SecondProcessView"),
```
buehlefs, line 70: Why is this starting with "second"?

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/routes.py
```diff
@@ -0,0 +1,361 @@
...
+        task.apply_async()
+
+        return redirect(
+            url_for("tasks-api.TaskView", task_id=str(db_task.id)), HTTPStatus.SEE_OTHER
+        )
+
+
+@PDPreprocessing_BLP.route("/<int:db_id>/<int:step_id>/ui/")
+class SecondMicroFrontend(MethodView):
+    """Micro frontend for the db manager plugin."""
```
buehlefs, line 198: copy pasta

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/routes.py
```diff
@@ -0,0 +1,361 @@
...
+            TASK_LOGGER.error(msg)
+            raise KeyError(msg)
+
+        db_task.parameters = SecondInputParametersSchema().dumps(arguments)
+
+        db_task.clear_previous_step()
+        db_task.save(commit=True)
+
+        # all tasks need to know about db id to load the db entry
+        task: chain = second_task.s(
```
buehlefs, line 298: The naming is quite confusing with "second_task" being used in "FinalProcess"...

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/routes.py
```diff
@@ -0,0 +1,361 @@
...
+        # save errors to db
+        task.link_error(save_task_error.s(db_id=db_task.id))
+        task.apply_async()
+
+        return redirect(
+            url_for("tasks-api.TaskView", task_id=str(db_task.id)), HTTPStatus.SEE_OTHER
+        )
+
+
+@PDPreprocessing_BLP.route("/<int:db_id>/<int:step_id>-process0/")
```
buehlefs, line 310: why is this ending with 0 and not `-process-second`
also inconsistent with the ui endpoint that separates the `step_id` with a `/`

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/routes.py
```diff
@@ -0,0 +1,361 @@
...
+        db_task.clear_previous_step()
+        db_task.save(commit=True)
+
+        # all tasks need to know about db id to load the db entry
+        task: chain = second_task.s(db_id=db_task.id, step_id=step_id) | add_step.s(
+            db_id=db_task.id,
+            step_id=next_step_id,
+            href=href,
+            ui_href=ui_href,
+            prog_value=50,
```
buehlefs, line 353: You could try to calculate a more usable progress value (i.e. `(nr_of_steps / (nr_of_steps + 1)) * 100`)
Or use a progress mode that is not based on percentages (may not be supported by the add_step task).

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/schemas.py
```diff
@@ -0,0 +1,271 @@
...
+        return str(self.__dict__)
+
+
+class FirstInputParametersSchema(FrontendFormBaseSchema):
+    file_url = FileUrl(
+        required=True,
+        allow_none=False,
+        metadata={
+            "label": "File URL",
+            "description": "A url to the file.",
```
buehlefs, line 56: Remove redundant descriptions if the label already communicates the same information (the description just uses screenspace without providing any benefits in this case)

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/schemas.py
```diff
@@ -0,0 +1,271 @@
...
+
+
+class SecondInputParametersSchema(FrontendFormBaseSchema):
+    preprocessing_enum = EnumField(
+        PreprocessingEnum,
+        required=True,
+        allow_none=False,
+        metadata={
+            "label": "Preprocessing Type",
+            "description": """Type of preprocessing the file will undergo.\n
```
buehlefs, line 97: Three `"""` should already keep line breaks as is without the `\n` escape

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/schemas.py
```diff
@@ -0,0 +1,271 @@
...
+        },
+    )
+    keep = EnumField(
+        KeepEnum,
+        required=False,
+        allow_none=True,
+        metadata={
+            "label": "Keep",
+            "description": """Determines which duplicates (if any) to keep.\n
+            - first: Drop duplicates except for the first occurrence.\n
```
buehlefs, line 155: This might trip up the markdown rendering as `"""` does not strip leading indentation.

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/tasks.py
```diff
@@ -0,0 +1,146 @@
...
+    df = read_csv(file_url)
+
+    # Output data
+    with SpooledTemporaryFile(mode="w") as output:
+        df.to_csv(output, index=False)
+        task_file = STORE.persist_task_result(
+            db_id,
+            output,
+            "file.csv",
+            "entity",
```
buehlefs, line 72: You might want to keep the original data type as is (but It might be hard to get that information here...). Just add a TODO comment here

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/tasks.py
```diff
@@ -0,0 +1,146 @@
...
+            "text/csv",
+        )
+        task_data.data["file_url"] = task_file.file_storage_data
+        # task_data.data["pandas_html"] = df.to_html(max_rows=100)
+        task_data.data["pandas_html"] = build_table(df, "grey_light")
```
buehlefs, line 77: This might get larger than what a single row in the Database can store. Maybe use the blob storage table instead.

buehlefs (changes requested):

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/templates/second_template.html
```diff
@@ -0,0 +1,252 @@
+{% extends "simple_template.html" %}
+
+{% block head %}
+    {{ super() }}
+{% endblock head %}
+
+{% block content %}
+<div id="pandas_html" style='height: 200px; overflow: auto; width: fit-content'>{{ pandas_html }}</div>
```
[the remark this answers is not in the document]
BenediktRiegel: There are two methods (that I know of) on how to include already rendered html in such a manner.
1. ``{{ pandas_html | safe }}`` (or a similare command, I am not sure)
2. Before passing ``pandas_html`` to the template, we wrap it in a Markup object, i.e. we pass it like this ``Markup(pandas_html)``.

I am using the second option for ``pandas_html``, ``columns_and_rows_html`` and ``columns_as_optionlist``. I hope this helps.

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/templates/second_template.html
```diff
@@ -0,0 +1,252 @@
...
+            }
+        }
+
+        function axis_change() {
+            set_subset_checklist(axis_val.value);
+            save_subset_checklist();
+        }
+
+        function preprocessing_enum_change() {
+            axis_dis.style.display = "none";
```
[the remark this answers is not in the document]
BenediktRiegel: Thanks! 
I couldn't make out a difference for "hidden" and "display = none" in this use case. Should this be changed or may it stay like this?

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/templates/second_template.html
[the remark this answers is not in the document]
buehlefs: Yeah, I did only see the python side after writing this comment, so this should be ok.

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/templates/second_template.html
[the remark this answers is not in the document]
buehlefs: It can stay like this.

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/tasks.py
BenediktRiegel: This is a rendered html of the retrieved table. It will be displayed on the MicroFrontend, to help the user decide which preprocessing step he wants to take next. If this is too big to save "normaly" within the database, it will probably also be too big, to render on the MicroFrontend and still be of use to a user, right?
Maybe instead of using the blob format, the displayed amount of the table should be limited and thus, avoiding a big rendered html code. What do you think?

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/backend/checkbox_list.py
```diff
@@ -0,0 +1,13 @@
+from flask.templating import render_template
+
+
+def get_checkbox_list_dict(dictionary: dict):
```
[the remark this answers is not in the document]
buehlefs: it is fine as is

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/tasks.py
buehlefs: That could also work. But make sure to display an error message to the user in case the preview was too large.

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/routes.py
buehlefs: Only the naming is confusing here. Naming something purely by the order it appears in (first_, second_, final_) works, but usually does not convey what the individual tasks do. Something like `setup`, `preprocessing`, `final/finishing`  would be less confusing. (good names are hard)

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/tasks.py
```diff
@@ -81,16 +80,15 @@ def first_task(self, db_id: int) -> str:
...
+        task_data.data["columns_list"] = df.columns.tolist()
 
     task_data.save(commit=True)
 
     return "Saved original csv file"
 
 
 @CELERY.task(name=f"{PDPreprocessing.instance.identifier}.second_task", bind=True)
-def second_task(self, db_id: int, step_id: int) -> str:
+def preprocessing_task(self, db_id: int, step_id: int) -> str:
```
buehlefs, line 91: also change name above (line 90)

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/templates/pd_preprocessing_step_2.html
```diff
@@ -2,12 +2,30 @@
 
 {% block head %}
     {{ super() }}
+    <style>
+        ul.qhana-form-input {
+            max-height: 150px;
```
buehlefs, line 7: I am not sure about the max height and hidden overflow. This will just clip long entries, which is bad for usability. It is usually better to have a UI that looks bad but at least works (and in this case: shows everything) in such edge cases instead of one that always looks great but is unusable in some cases.

buehlefs (commented):
Just a few minor issues left

buehlefs:
After trying out the plugin in the QHAna UI:
The styles look ok (though the preview table does not quite work in dark mode).

However, the file url input is missing the "choose data" button (and the plugin metadata does not contain any input specifications). Both should be fixed.

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/tasks.py
BenediktRiegel: I can limit the number of rows and columns that will be included in the rendered html of the table, to avoid large storage usage. This doesn't garuantee that the rendered html will be small, though, e.g. if the dataframe has only one row and one column, with a realy big entry. Do you have any suggestions on how to deal with this?

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/templates/pd_preprocessing_step_2.html
BenediktRiegel: Is "overflow-wrap: break-word;" also a viable option?
It inserts line breaks, if the label of a checkbox is too long. Though, this does not ensure that the whole word is visible at once, but maybe this is a good compromise between usability and looks. What do you think?

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/tasks.py
buehlefs: You could check the html length before comitting to the DB and if the HTML is too large commit a fallback html instead. Or be creative with shortening values. But generally do not put too much work into that safety feature.

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/templates/pd_preprocessing_step_2.html
buehlefs: Allowing breaks inside a word is better, but only if the height is not limited (or scrolls on overflow). Otherwise, part of the label will still be hidden in the overflow.

buehlefs: With overflow auto in one direction it should be ok (I have missed the `auto` after `hidden` in the css)

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/tasks.py
BenediktRiegel: I have updated this process. Hopefully, it suffices.

stable_plugins/classical_ml/data_preparation/pandas_preprocessing/routes.py
```diff
@@ -64,10 +65,19 @@ def get(self):
             entry_point=EntryPoint(
                 href=url_for(f"{PDPreprocessing_BLP.name}.FirstProcessView"),
                 ui_href=url_for(f"{PDPreprocessing_BLP.name}.FirstMicroFrontend"),
-                data_input=[],
+                data_input=[
+                    InputDataMetadata(
+                        data_type="entity",  # TODO Allow any data type as input
```
buehlefs, line 70: use `"*"` or `"*/*"` for that if you want to allow anything

buehlefs (approved):

buehlefs (approved):

PR #104, opened 2023-09-13
Refactor json preview for better performance

buehlefs:


infacc (approved):
I recognized 2 problems with the JSON visualization plugin while testing:
- The site hangs for quite a while when using large JSON files (~25 MB)
- The submit button (expand "Visualization Options") causes an error (I'll have to investigate this further to make sure it's not caused by my dev setup).

However, both issues seem to exist in the main branch as well.

buehlefs:
With (most) visualization plugins, the submit button is not really expected to work (as there is no data transformation required). The visualization happens entirely client side for the JSON plugin. (So yes, the error is expected, but I could probably just hide the submit button...)

Visualizing very large files (anything >4mb) will always put a strain on the browser, even just for pure text. The performance should be noticeably better now (compared with before) when visualizing not quite as large JSON files (e.g., files from the data creator with >500 points). For very large files visualization plugins may decide to only display a fragment of the data, but this is not implemented yet.

PR #106, opened 2023-10-09
Fix/confidential log

infacc:
- Change `input_type` of IBMQ tokens from `"text"` to `"password"` (in plugin input parameter schemas)
- Hide IBMQ token from log


qhana_plugin_runner/util/logging.py
```diff
@@ -22,3 +22,18 @@ def get_logger(app: Flask, name: str) -> Logger:
     """Utitlity method to get a specific logger that is a child logger of the app.logger."""
     logger_name = f"{app.import_name}.{name}"
     return getLogger(logger_name)
+
+
+def make_log_data(data: dict, private_fields: tuple = ("ibmq_token", "db_password")):
```
buehlefs, line 27:
```suggestion
def redact_log_data(data: dict, private_fields: tuple = ("ibmq_token", "db_password")):
```

buehlefs (approved):

PR #108, opened 2023-10-16
Optimizer Plugin

ProbstenHias:
##  Implementing Variational Quantum Algorithms as Compositions of Reusable Microservice-based Plugins
by Matthias Weilinger

This is the code for the Master Thesis with the title `Implementing Variational Quantum Algorithms as Compositions of Reusable Microservice-based Plugins`

It implements an framework to allow for optimization by coordinating the interaction of several plugins that together minimize an objective function.
By doing so it creates a environment that allows for future implementations of Variational Quantum Algorithms.
Furthermore the developed system for plugin interaction can be used in any other scenario where plugins should communicate with each other.

The thesis compares this implementation to an alternative approach that turned out to be less modular and performant.
This second approach can be found in #109.


plugins/optimizer/objective_functions/neural_network/schemas.py
```diff
@@ -0,0 +1,40 @@
...
+    number_of_neurons: int
+
+
+class HyperparamterInputSchema(FrontendFormBaseSchema):
+    number_of_neurons = ma.fields.Integer(
+        required=True,
+        allow_none=False,
+        metadata={
+            "label": "Number of neurons",
+            "description": "Number of neurons for the neural network.",
```
buehlefs, line 33: Layout of the neural network?

plugins/optimizer/minimizer/scipy_minimizer/tasks.py
```diff
@@ -0,0 +1,176 @@
...
+        minimize_params["jac"] = jac
+
+    if minimizer_input_data.calc_loss_and_gradient_endpoint_url:
+        loss_and_jac = loss_and_jac_(
+            minimizer_input_data.calc_loss_and_gradient_endpoint_url
+        )
+        minimize_params["jac"] = True
+        minimize_params["fun"] = loss_and_jac
+
+    result = scipy_minimize(**minimize_params)
```
buehlefs, line 161: Investigate whether this can be paused or broken into multiple minimization steps.

buehlefs (commented):
We will need to fix a few things before merging this PR. Critical issues that need to be addressed before merging:

* absolute imports in plugins
* shared code of plugins outside the plugin runner
* plugin dependencies specified in plugin runner dependencies

The rest can be addressed in subsequent PRs.

We will implement the fixes ourselves.

PR #110, opened 2023-11-03
Feature: qasm visualization plugin

infacc:
The plugin works as follows: If a URL to a QASM file is provided, the plugin generates an image from it and stores the image. When the same URL is requested again, it provides the stored image.


infacc:
TODO: Currently, the image generation isn't done by the worker (celery task), which may block the API during the image generation.

stable_plugins/visualization/file_types/qasm_visualization_templates/qasm_visualization.html
```diff
@@ -0,0 +1,102 @@
...
+        function getUrl() {
+            return document.querySelector('input[name=data]').value;
+        }
+
+        function updatePreview() {
+            const circuitUrl = getUrl()
+            if (currentUrl !== circuitUrl) {
+                document.querySelector('#spinner').style.display = '';
+                const preview = document.querySelector('#circuit-image');
+                const url = "{{ url_for(request.endpoint)[:-3] }}circuits?circuit-url=" + circuitUrl;
```
buehlefs, line 75: You can just calculate the correct URL in the python code and pass it as a context var or hardcode the correct endpoint name in the template. Either way, do not string concat together parts for the url but pass everything (including query params) to the `url_for` function:
`url_for('<blueprint-name>.<endpoint-name>`, **{'circuit-url': circuitUrl})`

(Dictionary unpacking `**` is required for url params with `-`)
The correct endpoint name can be found using the `/debug/routes` endpoint that is available in the dev setup.

stable_plugins/visualization/file_types/qasm_visualization_templates/qasm_visualization.html
```diff
@@ -0,0 +1,102 @@
...
+            {% call forms.render_form(target="microfrontend") %}
+            {{ forms.render_fields(schema, values=values, errors=errors) }}
+            <div class="qhana-form-buttons">
+                {{ forms.submit("validate", target="microfrontend") }}
+            </div>
+            {% endcall %}
+        </details>
+    </div>
+
+    <img id="circuit-image" class="qhana-data-preview" onload="onCircuitImageLoad()" onerror="onCircuitImageError()">
```
buehlefs, line 58: Do not set event listeners like this in HTML! This can be a security risk so webpages can be configured to disallow this. Instead, use `<element>.addEventListener('load', onCircuitImageLoad)` from your JavaScript to add the listeners and `document.getElementById` or `document.querySelector` to find your element.

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -0,0 +1,252 @@
...
+        return send_file(path, mimetype="image/png")
+    try:
+        with open_url(circuit_url) as quasm_response:
+            circuit_qasm = quasm_response.text
+    except HTTPError:
+        TASK_LOGGER.info(f"Could not load circuit from {circuit_url}")
+        abort(HTTPStatus.NOT_FOUND)
+    circuit = QuantumCircuit.from_qasm_str(circuit_qasm)
+    fig = circuit.draw(output="mpl")
+    fig.savefig(path)
```
buehlefs, line 190: You can do this inside a task and just await the result with a timeout of <10s in this response. In case of a timeout error respond wit the appropriate error code (and maybe the browser will try again later to see if the image is available automatically).

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -0,0 +1,252 @@
...
+
+    def get_api_blueprint(self):
+        return QASM_BLP
+
+
+TASK_LOGGER = get_task_logger(__name__)
+
+
+@CELERY.task(name=f"{QasmVisualization.instance.identifier}.demo_task", bind=True)
+def demo_task(self, db_id: int) -> str:
```
buehlefs, line 243: If you want to do the work (which would not be too much in your case) this task could also calculate the circuit visualization and then output it like any other processing plugin task would to the file store. This way, the plugin could also be used as a processing plugin. (This is harder to do for plugins with only client side visualization, which is why the existing plugins do not do this yet.)

poetry.lock
buehlefs, line 1: Revert changes to this file

buehlefs (changes requested):

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -0,0 +1,252 @@
...
+            )
+        )
+
+
+@QASM_BLP.route("/circuits/")
+def get_circuit_image():
+    circuit_url = request.args.get("circuit-url", None)
+    if circuit_url is None:
+        abort(HTTPStatus.NOT_FOUND)
+    filename = hashlib.md5(circuit_url.encode("utf-8")).hexdigest() + ".png"
```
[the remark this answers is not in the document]
infacc: I changed the hash to `sha256`. Currently, only the hash of the file URL is hashed. Does it make more sense to hash the file contents to distinguish different circuits?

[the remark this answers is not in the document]
buehlefs: not really, in QHAna every file version gets its own unique URL. For outside data it should be enough to clear cache entries that are older than x hours or days and regenerate them instead. (This would require storing a time...)

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -0,0 +1,252 @@
...
+    path = pathlib.Path(__file__).parent.absolute() / "img" / filename
```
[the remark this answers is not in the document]
infacc: Committing a `DataBlob` without explicitly setting the `id` doesn't seem to work:

```python
DataBlob(plugin_id='plugin-id', value=<binary data>)
DB.session.commit()
```

<details>
<summary>Example error</summary>

```
Traceback (most recent call last):
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/flask/app.py", line 2190, in wsgi_app
    response = self.full_dispatch_request()
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/flask/app.py", line 1486, in full_dispatch_request
    rv = self.handle_user_exception(e)
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/flask_cors/extension.py", line 165, in wrapped_function
    return cors_after_request(app.make_response(f(*args, **kwargs)))
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/flask/app.py", line 1484, in full_dispatch_request
    rv = self.dispatch_request()
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/flask/app.py", line 1469, in dispatch_request
    return self.ensure_sync(self.view_functions[rule.endpoint])(**view_args)
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/flask_smorest/response.py", line 89, in wrapper
    func(*args, **kwargs)
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/webargs/core.py", line 594, in wrapper
    return func(*args, **kwargs)
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/flask_smorest/arguments.py", line 82, in wrapper
    return func(*f_args, **f_kwargs)
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/qhana_plugin_runner/api/jwt_helper.py", line 96, in wrapper
    return func(*args, **kwargs)
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/stable_plugins/visualization/file_types/qasm_visualization.py", line 207, in get_circuit_image
    blob = DataBlob.set_blob(
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/qhana_plugin_runner/db/models/virtual_plugins.py", line 316, in set_blob
    DB.session.commit()
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/sqlalchemy/orm/scoping.py", line 553, in commit
    return self._proxied.commit()
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 1906, in commit
    trans.commit(_to_root=True)
  File "<string>", line 2, in commit
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/sqlalchemy/orm/state_changes.py", line 103, in _go
    self._raise_for_prerequisite_state(fn.__name__, current_state)
  File "/mnt/e/Temp/hiwi/qhana-plugin-runner/.venv/lib/python3.10/site-packages/sqlalchemy/orm/session.py", line 906, in _raise_for_prerequisite_state
    raise sa_exc.PendingRollbackError(
sqlalchemy.exc.PendingRollbackError: This Session's transaction has been rolled back due to a previous exception during flush. To begin a new transaction with this Session, first issue Session.rollback(). Original exception was: (sqlite3.IntegrityError) NOT NULL constraint failed: DataBlob.id
[SQL: INSERT INTO "DataBlob" (plugin_id, value) VALUES (?, ?)]
[parameters: ('qasm-visualization@v0-2-0', <memory at 0x7f37ee002980>)]
(Background on this error at: https://sqlalche.me/e/20/gkpj) (Background on this error at: https://sqlalche.me/e/20/7s2a)
```
</details>

I suppose this is because `DataBlob` has a multi-column primary key:

https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/24e02eaa1d36daeede9d57c61a0d5f53c8fe3f28/qhana_plugin_runner/db/models/virtual_plugins.py#L267-L273

Would it make sense to replace `DataBlob.id` (int) by `DataBlob.key` (str) such that we can reference DataBlobs by `plugin_id` and `key` pairs like for PluginStates?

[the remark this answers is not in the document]
buehlefs: That may be because the plugin id is also part of the primary key and this prevents autoincrement from working. Yes, it is probably for the best to align this with the plugin state and change the id to a key column.

stable_plugins/visualization/file_types/qasm_visualization.py
infacc: The `qasm-visualization` plugin can now be used as processing and visualization plugin. Images are stored as files (like the outputs of other processing plugins) and a mapping between `url_hash`es ans file paths is stored in the PluginState. Images are created through a celery task. If an image is not yet created, flask responds with status code `503 Service Unavailable` and the HTTP header [`Retry-After`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Retry-After) set. Since the `Retry-After` header doesn't work for all browsers, the "timeout before retry" is implemented in [`qasm-visualization.html`](https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/94d112c27f1f1f083334afa0fbdc7d0011e2a5bb/stable_plugins/visualization/file_types/qasm_visualization_templates/qasm_visualization.html#L106-L124).

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -249,4 +287,45 @@ def demo_task(self, db_id: int) -> str:
...
+            db_id,
+            output,
+            f"circuit_{url_hash}.png",
+            "image/png",
+            "image/png",
+        )
+        PluginState.set_value(
+            QasmVisualization.instance.identifier,
+            url_hash,
+            file.file_storage_data,
```
buehlefs, line 323: This is not safe to do if another file store is configured as default!

buehlefs (changes requested):
The interactions with the file store are not safe right now. I suggest the following refactoring (outlined in pseudocode python):

Celery Tasks:
```python
@CELERY.task(...)
def gen_image(url, hash, plugin_id):
    # generate image
    DataBlob.save(plugin_id, hash, generated_image)
    PluginData.remove(plugin_id, hash, commit=True)


@CELERY.task(..., retry_on=ImageNotFinishedError)
def process(db_id, url, hash, plugin_id):
    if not (image := DataBlob.get(plugin_id, hash)):  # image not already in DB
        if not (task_id := PluginData.get(plugin_id, hash)):  # no task generating image started
            task_result = gen_image.s(url, hash, plugin_id).apply_async()
            PluginData.save(plugin_id, hash, task_result.id, commit=True)
        raise ImageNotFinishedError()  # auto retry with auto backoff is probably easiest to implement (see workflow plugin for examples)
    # save image to task file store (can probably use BytesIO instead of spooledTemporaryFile)
    STORE.save(...)
```

Routes:

```python
# /circuits/?url=...
def get(url):
    if not (image := DataBlob.get(plugin_id, hash)):
        if not (task_id := PluginData.get(plugin_id, hash)):
            task_result = gen_image.s(url, hash, plugin_id).apply_async()
            PluginData.save(plugin_id, hash, task_result.id, commit=True)
        # optional:
        image_status = task_result.get(timeout=5) # return image directly if positive
        return HTTPStatus.NOT_READY_YET
    return image  # always serve image from DataBlob table here

# /processing/
def post(url):
    # standard processing task
    process.s(...).apply_async()
    ...
```

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -172,47 +177,49 @@ def render(self, data: Mapping, errors: dict, valid: bool):
...
-            start_task(circuit_url)
-            timeout = 2
-        else:
-            # increase timeout for next request
+    url_hash = hashlib.sha256(url.encode("utf-8")).hexdigest()
+    image = DataBlob.get_value(QasmVisualization.instance.identifier, url_hash, None)
+    if image is None:
+        response = Response("Image not yet created!", HTTPStatus.SERVICE_UNAVAILABLE)
+        response.headers["Retry-After"] = 5
+        if not PluginState.get_value(
```
buehlefs, line 201: In the else case you could get an awaitable asyncResult from the task id. (see task api in plugin runner)

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -275,57 +276,61 @@ def get_requirements(self) -> str:
...
             QasmVisualization.instance.identifier,
-            url_hash,
+            hash,
             "",
             commit=True,
         )
-        PluginState.delete_value(
-            QasmVisualization.instance.identifier, f"timeout_{url_hash}", commit=True
-        )
+        PluginState.delete_value(QasmVisualization.instance.identifier, hash, commit=True)
```
buehlefs, line 293: Double commit (line 291)

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -275,57 +276,61 @@ def get_requirements(self) -> str:
...
-            QasmVisualization.instance.identifier,
-            url_hash,
-            file.file_storage_data,
-            commit=True,
-        )
-
-    PluginState.delete_value(
-        QasmVisualization.instance.identifier, f"timeout_{url_hash}", commit=True
+    figfile = BytesIO()
+    fig.savefig(figfile, format="png")
```
buehlefs, line 299: This should be save for images, but a SpooledTemporaryFile has the advantage, that it evicts the content from ram to disk if it gets too large automatically. (Again this is probably only relevant for stupidly large images of stupidly large circuits, so no need to fix)

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -275,57 +276,61 @@ def get_requirements(self) -> str:
...
+    figfile.seek(0)
+    DataBlob.set_value(
+        QasmVisualization.instance.identifier, hash, figfile.getvalue(), commit=True
     )
+    TASK_LOGGER.info(f"Stored image of circuit {circuit.name}.")
+    PluginState.delete_value(QasmVisualization.instance.identifier, hash, commit=True)
```
buehlefs, line 305: Double commit (line 302)

buehlefs (commented):
Looks good. Just some minor comments left

buehlefs:
About the used status code: 202 Accepted looks like it would fit better when the image is not created yet. If we need javascript anyway we can use the more standard complient code.

stable_plugins/visualization/file_types/qasm_visualization.py
```diff
@@ -196,10 +196,10 @@ def get_circuit_image(data: Mapping):
     url_hash = hashlib.sha256(url.encode("utf-8")).hexdigest()
     image = DataBlob.get_value(QasmVisualization.instance.identifier, url_hash, None)
     if image is None:
-        response = Response("Image not yet created!", HTTPStatus.SERVICE_UNAVAILABLE)
-        response.headers["Retry-After"] = 5
-        if not PluginState.get_value(
-            QasmVisualization.instance.identifier, url_hash, None
+        if not (
```
buehlefs, line 199: To avoid such nesting, use a standard variable before the if (however this is still ok, so no need to change)

buehlefs (approved):

PR #112, opened 2023-11-22
Workflow editor plugin

buehlefs:
- [x] integrate webcomponent in microfrontend
- [x] check styles when loaded in qhana
- [x] add custom interactions and config (e.g. for file opening/saving)
- [x] make sure everything works (save. load, transform, deploy, etc.)
- [x] Add management UI for saved workflows, maybe in the workflow management plugin (then using the API to fetch and delete workflows...)


PhilWun (approved):
lgtm

PR #118, opened 2024-01-16
circuit demo plugin

buehlefs:
Adds a small plugin to demonstrate how circuit executor plugins can be used by other plugins.

TODO:
- [x] Implement circuits for all bell states
- [x] Fix outputs in plugin metadata
- [x] Add more outputs?
- [ ] Add nicer result and circuit visualization?


PhilWun (approved):
lgtm

PR #126, opened 2024-10-04
Fix consistency issues of plugin tags

buehlefs:
- [x] Automate plugin documentation to better spot inconsistencies
- [x] Fix tags in stable plugins and bump plugin versions
- [x] Fix issues with plugin identifiers


docs/plugin_autodoc.py
```diff
@@ -0,0 +1,176 @@
...
+                f"|{data_input['dataType']}|{', '.join(data_input['contentType'])}|{'✓' if data_input['required'] else '🗙'}|\n"
+            )
+        doc.write("\n\n")
+    if p["output"]:
+        doc.write("**Outputs:**\n\n")
+        doc.write("| Data Type | Content Type | Always |\n")
+        doc.write("|-----------|--------------| :----: |\n")
+        for data_output in p["output"]:
+            doc.write(
+                f"|{data_output['dataType']}|{', '.join(data_output['contentType'])}|{'✓' if data_output['required'] else '🗙'}|\n"
```
PhilWun, line 135: '🗙' is rendered as an unknown character in Firefox, Chrome, PyCharm etc. on MacOS.

PhilWun (changes requested):

plugins/optimization_coordinator/__init__.py
```diff
@@ -20,7 +20,7 @@
 from qhana_plugin_runner.util.plugins import QHAnaPluginBase, plugin_identifier
 
 _plugin_name = "optimizer"
-__version__ = "v1.0.0"
+__version__ = "v1.0.1"
```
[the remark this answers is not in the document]
buehlefs: I have just increased the version numbers for any plugin where I have changed something related to the metadata. It was not strictly necessary, but easier than verifying that some plugins did not need version number bumps...

docs/plugin_autodoc.py
buehlefs: Since this is a Unicode character, it depends on what font is installed (and used) to display this character. Can you suggest an alternative that renders nicely on macOS? (in the worst case we could use 'X' or even 'yes' 'no'...)

PhilWun (approved):
lgtm

PR #128, opened 2024-11-12
Prepare muse4music plugin for promotion to stable plugins

buehlefs:
- [x] remove test code
- [x] add ui form entry to pass muse4music URL as parameter
- [ ] move plugin to stable plugins

### DIFF
--- a/plugins/muse_for_music/routes.py
+++ b/plugins/muse_for_music/routes.py
@@ -122,8 +122,7 @@ class MicroFrontend(MethodView):
         data_dict = dict(data)
         # define default values
         default_values = {
-            "username": "admin",  # FIXME remove default later!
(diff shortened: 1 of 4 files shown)


plugins/muse_for_music/routes.py
```diff
@@ -122,8 +122,7 @@ def render(self, data: Mapping, errors: dict):
         data_dict = dict(data)
         # define default values
         default_values = {
-            "username": "admin",  # FIXME remove default later!
-            "password": "admin",  # FIXME remove default later!
+            "url": "",
```
PhilWun, line 125: shouldn't this be "muse_url"?

buehlefs: Fixed, it was `museUrl`

PR #129, opened 2024-11-13
Add test to find bad imports in plugins

buehlefs:


PhilWun (changes requested):
- When importing an external dependency (e.g. qiskit) in the __init__.py file of a multi-file plugin, the error message is "Plugins must use relative imports!" instead of "Found early import of external dependency".
- When using absolute import to import a plugin file (e.g. schemas.py) inside a multi-file plugin (e.g. in tasks.py), no error is raised

PhilWun (approved):
lgtm

PR #137, opened 2025-03-07
Use UI href where possible for muse4music entities

buehlefs:


PhilWun (changes requested):
PersonEntity is missing the UI href

buehlefs:
Unfortunately the UI does not address individual people by URL (yet) so there is little use in having a UI href for people right now...

PhilWun (approved):
lgtm

PR #138, opened 2025-03-11
Improve data handling in Wu Palmer and Sym Max Mean plugins

PhilWun:
- set similarity of empty / missing attributes to 0
- use attribute metadata to fix list attribute parsing in Wu Palmer plugin for CSV files

### DIFF
--- a/stable_plugins/classical_ml/data_preparation/sym_max_mean.py
+++ b/stable_plugins/classical_ml/data_preparation/sym_max_mean.py
@@ -354,6 +354,8 @@ def calculation_task(self, db_id: int) -> str:
 
                 if ent_attr1 is None or ent_attr2 is None:
                     sym_max_mean = None  # TODO: add handling of missing values
+                elif len(ent_attr1) == 0 or len(ent_attr2) == 0:
+                    sym_max_mean = None
                 else:
                     if isinstance(ent_attr1, set):
                         ent_attr1 = list(ent_attr1)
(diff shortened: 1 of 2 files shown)


stable_plugins/classical_ml/data_preparation/sym_max_mean.py
```diff
@@ -357,47 +355,56 @@ def calculation_task(self, db_id: int) -> str:
...
+
                     if not isinstance(ent_attr1, list):
                         ent_attr1 = [ent_attr1]
 
                     if not isinstance(ent_attr2, list):
                         ent_attr2 = [ent_attr2]
 
-                    # calculate Sym Max Mean
+                    if len(ent_attr1) == 0 or len(ent_attr2) == 0:
+                        sym_max_mean = 0
```
buehlefs, line 371: Should this also apply if one side is None? (and if both are None/empty, should they have a similarity > 0 or even 1?) Maybe those default values could even be parameters in the frontend.

stable_plugins/classical_ml/data_preparation/wu_palmer.py
```diff
@@ -417,8 +424,8 @@ def add_similarities_for_entities(
 
     for val1 in values1:
         for val2 in values2:
-            if val1 is None or val2 is None:
-                sim = None
+            if val1 is None or val1 == "" or val2 is None or val2 == "":
```
buehlefs, line 427: move check for val 1 to outer loop and maybe combine None and empty checks into Falsy checks:

```python
if not val1:  # catches None and empty strings
    continue
```

Also, don't use an else after a continue or return in an if

stable_plugins/classical_ml/data_preparation/wu_palmer.py
```diff
@@ -490,6 +497,16 @@ def calculation_task(self, db_id: int) -> str:
         attributes,
         root_has_meaning_in_taxonomy,
     ) = load_input_parameters(db_id)
+    deserializer: Callable[[tuple[str, ...]], tuple[any, ...]] | None = None
+
+    with open_url(entities_metadata_url) as entities_metadata_file:
+        entities_metadata_list = list(
+            load_entities(entities_metadata_file, "application/json")
```
buehlefs, line 504: use the get_mimetype function to guess the mimetype here instead of hardcoding json

stable_plugins/classical_ml/data_preparation/wu_palmer.py
```diff
@@ -498,16 +515,19 @@ def calculation_task(self, db_id: int) -> str:
 
         for ent in load_entities(entities_data, mimetype):
             if hasattr(ent, "_asdict"):  # is NamedTuple
+                ent_attributes: tuple[str, ...] = ent._fields
+                ent_tuple = type(ent)
```
buehlefs, line 519: maybe move this into the inner if deserializer is None

and please copy the logic from ensure_array for the `isinstance(item, EntityTupleMixin)` case

buehlefs (changes requested):

stable_plugins/classical_ml/data_preparation/wu_palmer.py
```diff
@@ -514,17 +516,16 @@ def calculation_task(self, db_id: int) -> str:
         entities = []
 
         for ent in load_entities(entities_data, mimetype):
-            if hasattr(ent, "_asdict"):  # is NamedTuple
-                ent_attributes: tuple[str, ...] = ent._fields
-                ent_tuple = type(ent)
-
+            if isinstance(ent, EntityTupleMixin):  # is NamedTuple
                 if deserializer is None:
+                    ent_attributes: tuple[str, ...] = ent._fields
```
buehlefs, line 521: maybe use `type(ent).entity_attributes` instead

stable_plugins/classical_ml/data_preparation/sym_max_mean.py
buehlefs: If both are empty/None, then None is the correct value, if only one side is empty/None, than 0 should be safe (they are not similar in that case).
How is None handled by the next in line plugin? Maybe that could reasonably assume None==0?

PhilWun: The "similarity to distances" plugin currently doesn't handle None values. I think it should just output None as well. The plugin after that would be the aggregator. I think it makes sense for this plugin to handle None values.

buehlefs (approved):

PR #139, opened 2025-03-14
Data Join Plugin

buehlefs:
This will be a multi-step plugin that allows joining additional attributes to entities.

Steps:
1. Select base entities (the entities to join data against)
2. Select entities to join to base & select against which attributes to match their ID value [possibly a repeatable step]
3. Maybe: select which attributes to actually keep and discard the rest

The plugin should output the joined entity data and the relevant attribute metadata


PhilWun (approved):
lgtm

PR #140, opened 2025-03-18
Improve handling of None values in MUSE plugins

PhilWun:
- similarities to distances converter
  - a None similarity will result in a None distance
- aggregator
  - the user can choose how to handle None attribute values
  - ignore: None values are removed and only the not None values are used for the aggregation
  - mean: None values are replaced by the mean distance of the respective attribute
  - max: None values are replaced by the maximum distance of the respective attribute

### DIFF
--- a/stable_plugins/classical_ml/data_preparation/aggregators.py
+++ b/stable_plugins/classical_ml/data_preparation/aggregators.py
@@ -116,7 +116,10 @@ class InputParametersSchema(FrontendFormBaseSchema):
         required=True,
         metadata={
             "label": "Missing data handling",
-            "description": "Defines how a missing attribute distance should be handled.",
+            "description": """Defines how a missing attribute distance should be handled.
+- ignore: null values are removed and only the not null values are used for the aggregation
+- mean: null values are replaced by the mean distance of the respective attribute
+- max: null values are replaced by the maximum distance of the respective attribute""",
             "input_type": "select",
         },
     )
--- a/stable_plugins/classical_ml/data_preparation/aggregators.py
+++ b/stable_plugins/classical_ml/data_preparation/aggregators.py
@@ -301,6 +301,10 @@ def calculation_task(self, db_id: int) -> str:
         attr_name = file_name[:-5]
 
         loaded_distances = json.load(file)
+        # FIXME: handle all distances being None
+        distances_without_none: list[float] = [
+            dist["distance"] for dist in loaded_distances if dist["distance"] is not None
+        ]
 
         if missing_data_handling == MissingDataHandling.ignore:
             # removes elements with None distance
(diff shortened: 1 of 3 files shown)


stable_plugins/classical_ml/data_preparation/aggregators.py
```diff
@@ -103,6 +111,15 @@ class InputParametersSchema(FrontendFormBaseSchema):
             "input_type": "select",
         },
     )
+    missing_data_handling = EnumField(
+        MissingDataHandling,
+        required=True,
+        metadata={
+            "label": "Missing data handling",
+            "description": "Defines how a missing attribute distance should be handled.",
```
buehlefs, line 119: Add description from PR for the individual options. Markdown is supported.

stable_plugins/classical_ml/data_preparation/aggregators.py
```diff
@@ -276,7 +297,55 @@ def calculation_task(self, db_id: int) -> str:
...
         attr_name = file_name[:-5]
 
-        attribute_distances[attr_name] = json.load(file)
+        loaded_distances = json.load(file)
+
+        if missing_data_handling == MissingDataHandling.ignore:
+            # removes elements with None distance
+            loaded_distances = [
+                dist for dist in loaded_distances if dist["distance"] is not None
+            ]  # FIXME: handle all distances being None
```
buehlefs, line 306: FIXME can be removed now...

stable_plugins/classical_ml/data_preparation/aggregators.py
```diff
@@ -276,7 +297,55 @@ def calculation_task(self, db_id: int) -> str:
...
+            # replaces None distances with the mean distance
+            new_list = []
+
+            for dist in loaded_distances:
+                if dist["distance"] is None:
+                    dist["distance"] = mean_distance
+
+                new_list.append(dist)
+
+            loaded_distances = new_list
```
buehlefs, line 324: Why create a new list for this? It should be fine modifying/reusing the existing one.

stable_plugins/classical_ml/data_preparation/aggregators.py
```diff
@@ -276,7 +297,55 @@ def calculation_task(self, db_id: int) -> str:
...
+        elif missing_data_handling == MissingDataHandling.max:
+            distances = [
```
buehlefs, line 326: this one could be a generator expression `(...)` instead of `[...]` as it is only used once

buehlefs (changes requested):

stable_plugins/classical_ml/data_preparation/aggregators.py
PhilWun: I added these fixmes, because it would still be a problem if all distances of an attribute are None or if the distances for all attributes of an entity pair are None.

buehlefs: But those cases cannot be handled in this plugin. (Or at least I don't see how they could be)

buehlefs (approved):

PR #148, opened 2025-05-22
Add plugin connection graph for usability

Purpurax:
The plugin connection graph should help with visualizing plugin interactions. Dash is used to display the graph and the layout is created through the [Klay algorithm for cytoscape](https://github.com/cytoscape/cytoscape.js-klay) (MIT license).


FlorianSang (approved):
LGTM :)

docs/plugin-connection-graph/example-all-plugins.md
buehlefs, line 1: This is an example/test file to test the parsing code?

buehlefs (changes requested):
Looking at the actual implementation, I would not merge this (at least not as it is now). Parsing the (already autogenerated) markdown file to extract the info is ok for a one off (so this is perfectly fine for your paper), but will be a nightmare to maintain. A better way to get this data in python object form would be to do what <https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/main/docs/plugin_autodoc.py> does.
The PR is also missing a small "how to use" documentation for the code. This should live right next to the code in question.

PR #150, opened 2025-08-21
QNN training

PhilWun:


plugins/qnn/loop.py
```diff
@@ -0,0 +1,661 @@
...
+
+    task_data: Optional[ProcessingTask] = ProcessingTask.get_by_id(id_=db_id)
+    params = LoopParametersSchema().loads(task_data.parameters or "{}")
+    state_url: Optional[str] = params.get("state", None)
+    ansatz_url: Optional[str] = params.get("ansatz", None)
+
+    ansatz_qasm: str
+    with open_url(ansatz_url) as quasm_response:
+        ansatz_qasm = quasm_response.text
+        ansatz_circuit = parse(ansatz_qasm)
```
buehlefs, line 510: maybe also try to parse as qasm2 if qasm3 failed

buehlefs (changes requested):
Not all comments need to be addressed for merging this. At the minimum, all `eval` uses need to be replaced by something safe. Tags and renaming should also happen in the PR. Other comments are mostly for remembering what we might want to do (if there is time to do it).

PR #164, opened 2026-03-03
Add entity conversion plugin for csv to json and json to csv

buehlefs:
Plugin is complete ~but also entirely untested. Bugs need to be found and fixed before merging...~
It should be ok to merge, even if minor bugs might still be present.

### DIFF
--- a/stable_plugins/file_utils/entity_conversion.py
+++ b/stable_plugins/file_utils/entity_conversion.py
@@ -131,8 +131,8 @@ class CsvPluginView(MethodView):
             version=plugin.version,
             type=PluginType.conversion,
             entry_point=EntryPoint(
-                href=url_for(f"{CSV_TO_JSON_BLP.name}.{CsvProcessView}"),
-                ui_href=url_for(f"{CSV_TO_JSON_BLP.name}.{CsvMicroFrontend}"),
+                href=url_for(f"{CSV_TO_JSON_BLP.name}.{CsvProcessView.__name__}"),
+                ui_href=url_for(f"{CSV_TO_JSON_BLP.name}.{CsvMicroFrontend.__name__}"),
                 plugin_dependencies=[],
                 data_input=[
                     InputDataMetadata(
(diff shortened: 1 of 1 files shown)


stable_plugins/file_utils/entity_conversion.py
```diff
@@ -0,0 +1,517 @@
+# Copyright 2021 QHAna plugin runner contributors.
```
GeorgDaniel: Reminder change to 2026

stable_plugins/file_utils/entity_conversion.py
```diff
@@ -0,0 +1,517 @@
...
+        location="form",
+        required=False,
+    )
+    @JSON_TO_CSV_BLP.require_jwt("jwt", optional=True)
+    def post(self, errors):
+        """Return the micro frontend with prerendered inputs."""
+        return self.render(request.form, errors, not errors)
+
+    def render(self, data: Mapping, errors: dict, valid: bool):
+        plugin = CsvToJson.instance
```
GeorgDaniel:
```suggestion
        plugin = JsonToCsv.instance
```

stable_plugins/file_utils/entity_conversion.py
```diff
@@ -0,0 +1,517 @@
...
+    @JSON_TO_CSV_BLP.arguments(JSONInputSchema(unknown=EXCLUDE), location="form")
+    @JSON_TO_CSV_BLP.response(HTTPStatus.SEE_OTHER)
+    @JSON_TO_CSV_BLP.require_jwt("jwt", optional=True)
+    def post(self, arguments):
+        """Start the conversion task."""
+        db_task = ProcessingTask(task_name="convert_json", parameters=arguments["data"])
+        db_task.save(commit=True)
+
+        # all tasks need to know about db id to load the db entry
+        task: chain = convert_csv.s(db_id=db_task.id) | save_task_result.s(
```
GeorgDaniel:
```suggestion
        task: chain = convert_json.s(db_id=db_task.id) | save_task_result.s(
```

stable_plugins/file_utils/entity_conversion.py
```diff
@@ -0,0 +1,517 @@
...
+
+                    ent = deserializer(ent)
+                    yield ent.as_dict()
+                else:
+                    if not ent_attributes:
+                        ent_attributes = tuple(ent.keys())
+                    yield ent
+
+        entities = load()
+        first = next(entities)
```
GeorgDaniel: Just a detail, probably not necessary, but we do not check for empty.

Could do something like this

```
first = next(entities, None)
        if first is None:
            raise ValueError("The input file does not contain any entities to convert.")

```

GeorgDaniel (commented):
Quick Test with the suggestions and it works
After applying these, approved

PR #173, opened 2026-05-10
Example tests for data_creator plugin

maxschuele:
- Add tests for datasets, routes, and schemas.
- Schema now validates per-data-type fields (clusters/centers/std for clusters, n_classes for classification, etc.) and template only renders fields relevant to the selected data type.
- CI: Install plugin dependencies (only for plugins that have test files)

Authored with the help of Claude Opus 4.7.

Closes UST-QuAntiL/QHAna4MUSE-EnPro26#23

### DIFF
--- a/docs/testing.rst
+++ b/docs/testing.rst
@@ -104,9 +104,13 @@ The fixture is function-scoped, so each test gets a fresh database.
 The ``app`` and ``client`` fixtures
 ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
 
-``app`` is a module-scoped Flask application built via :py:func:`~qhana_plugin_runner.create_app` with the same in-memory SQLite configuration as ``task_data``. Plugin discovery runs as part of ``create_app``, so every blueprint declared under ``PLUGIN_FOLDERS`` is registered on the returned app. Module scope amortises the startup cost of plugin discovery across the test cases in a file. Use ``app`` whenever a test needs the full configured application, an application context, or :py:func:`flask.url_for` without a request context (the test configuration sets ``SERVER_NAME`` so ``url_for`` can build URLs outside a request).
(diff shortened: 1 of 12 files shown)


docs/testing.rst
```diff
@@ -101,6 +101,25 @@ The ``task_data`` fixture
 
 The fixture is function-scoped, so each test gets a fresh database.
 
+The ``app`` and ``client`` fixtures
+~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
+
+``app`` is a module-scoped Flask application built via :py:func:`~qhana_plugin_runner.create_app` with the same in-memory SQLite configuration as ``task_data``. Plugin discovery runs as part of ``create_app``, so every blueprint declared under ``PLUGIN_FOLDERS`` is registered on the returned app. Module scope amortises the startup cost of plugin discovery across the test cases in a file. Use ``app`` whenever a test needs the full configured application, an application context, or :py:func:`flask.url_for` without a request context (the test configuration sets ``SERVER_NAME`` so ``url_for`` can build URLs outside a request).
```
buehlefs: In most text based formats with git it is usually better to wrap text at some point. Either wrap at a hard column limit, or put each sentence into its own line. This will improve the diff/merge experience on edits dramatically.

pyproject.toml
```diff
@@ -65,6 +65,9 @@ target-version = ["py310"]
 line-length = 90
 include = '\.pyi?$'
 
+[tool.ruff]
+  line-length = 90
```
buehlefs, line 69: We might want to switch everything to ruff in the future.

docs/testing.rst
maxschuele: added line breaks at sentence end

pyproject.toml
maxschuele: I use ruff in my editor and it is compatible with black except that it defaults to line length 88.
Can remove this from the MR if you prefer.

buehlefs: Its better to leave it in.

buehlefs (approved):

PR #180, opened 2026-05-30
muse4music importer: import taxonomy mappings

maxschuele:
this is what the taxonomies json outputs look like with this change:
image
empty string if no maping, else the mapping values are presented as a string

Closes UST-QuAntiL/QHAna4MUSE-EnPro26#34


Simonandy28 (approved):
Looks good.
Related pull regest from muse4muse repo: https://github.com/Muster-Suchen-und-Erkennen/muse-for-music/pull/78

PR #182, opened 2026-06-04
Parse taxonomy mappings to array

maxschuele:
- [x] Import mappings as array of numbers
- [x] normalize mappings on import to same length (per taxonomy)
- [x] Document mapping field for taxonomy datatype

Closes UST-QuAntiL/QHAna4MUSE-EnPro26#37


Simonandy28 (approved):
Looks good

PR #183, opened 2026-06-08
One hot encoding and conversion plugin fixes

ChrissiWHG:


maxschuele:
-> UST-QuAntiL/qhana-plugin-runner#185

PR #184, opened 2026-06-13
write test for data and content type, remove application/csv and update doc

ChrissiWHG:


maxschuele:
UST-QuAntiL/QHAna4MUSE-EnPro26#48

qhana_plugin_runner/plugin_utils/types.py
```diff
@@ -0,0 +1,89 @@
+"""
```
maxschuele: License comment is missing

qhana_plugin_runner/plugin_utils/types.py
```diff
@@ -0,0 +1,89 @@
...
+
+from enum import Enum
+
+
+class AllowedDataTypesWithFormat(Enum):
+    """
+    Allowed data types of the following format: namespace/name.\\
+    Data types, that are allowed, but do not follow this format
+    Also contains some data types, that DO NOT follow this format and *.
```
maxschuele: I don't really understand what this comment is trying to say

qhana_plugin_runner/plugin_utils/types.py
```diff
@@ -0,0 +1,89 @@
+"""
+Auto-generated module for QHAna plugin data types and content types.
```
maxschuele: What does auto-generated mean here? New types will still have to be added manually to this file right?

tests/test_plugin_metadata_types.py
```diff
@@ -0,0 +1,152 @@
...
+                    continue
+            except NotImplementedError:
+                continue
+
+            try:
+                response = client.get(f"/plugins/{plugin.identifier}/")
+            except Exception:
+                continue  # plugin's metadata endpoint crashes
+            if response.status_code != HTTPStatus.OK:
+                continue
```
maxschuele: Test should fail or at least print a warning if there is no blueprint or the request fails.
Would probably be fine to print a warning for now and create a followup issue to fix the affected plugins.

tests/test_plugin_metadata_types.py
```diff
@@ -0,0 +1,152 @@
...
```
maxschuele: Same thing here

conftest.py
```diff
@@ -25,6 +26,15 @@
 from qhana_plugin_runner.db.cli import create_db_function
 from qhana_plugin_runner.db.models.tasks import ProcessingTask
 
+
+def _load_plugin_folders():
+    for line in Path(".flaskenv").read_text(encoding="utf-8").splitlines():
+        if line.startswith("PLUGIN_FOLDERS="):
+            value = line.split("=", 1)[1].strip()
+            return [f for f in value.split(":") if f]
+    return []
```
maxschuele: Can be simplified to one-liner:
```suggestion
[f for f in dotenv_values(".flaskenv")["PLUGIN_FOLDERS"].split(":") if f]
```

maxschuele (changes requested):

tests/test_plugin_metadata_types.py
```diff
@@ -0,0 +1,152 @@
+"""
```
maxschuele, line 15: Add license comment

tests/test_plugin_metadata_types.py
```diff
@@ -0,0 +1,152 @@
...
+            if response.status_code != HTTPStatus.OK:
+                continue
+
+            metadata = response.get_json()
+            assert metadata
+
+            entry_point = metadata.get("entryPoint", {})
+
+            # Check data_input types
+            for input_item in entry_point.get("dataInput", []):
```
maxschuele: Should fail if dataInput is not in the entry_point dict.
Same for dataOutput.

Or are there plugins without input/output?

ChrissiWHG: I suppose that there are none, because the test didn't fail if checking against it.

maxschuele (approved):
looks good

PR #185, opened 2026-06-14
One-hot encoding handle muse4music taxonomies, entities to json marshaling convert sets to lists

maxschuele:
- entities to json marshaling: convert sets to lists because sets cant be marshaled to json
- Updated one-hot encoding plugin to also handle muse4music taxonomies.

For muse taxonomies entities look like this:
```json
"entities": ["", "30ern", "70ern", "Alter", "Kleinkind", "Erwachsener", "50ern", "Senior", "60ern", "90ern", "80ern", "40ern", "Baby", "Jugendlicher", "20ern", "Kind"]
```
Where `""` is the root node.

For muse4music the entities look like this:
```json
"entities": [{"ID": "t_Akkord_2", "tax_item_name": "root", "description": "", "mapping": [], "mapping_raw": ""}, {"ID": "t_Akkord_na", "tax_item_name": "na", "description": "", "mapping": [], "mapping_raw": ""}, ...]
```

These changes handle the different formats of the entities field and filter out the root nodes for both muse and muse4music taxonomies

Example mapping files:
[t_AmbitusEntwicklung.json](https://github.com/user-attachments/files/28927480/t_AmbitusEntwicklung.json)
[alterseindruck.json](https://github.com/user-attachments/files/28927499/alterseindruck.json)


stable_plugins/classical_ml/data_preparation/one-hot_encoding.py
```diff
@@ -360,38 +358,47 @@ def taxonomy_node_to_parent(taxonomy):
...
     if attribute in ancestor_nodes_dict:
         return ancestor_nodes_dict[attribute]
     else:
         parent = parent_node_dict[attribute]
-        result = get_ancestor_nodes(parent_node_dict, parent, ancestor_nodes_dict).copy()
-        if parent != "":
+        result = get_ancestor_nodes(
+            parent_node_dict, parent, ancestor_nodes_dict, root_id
+        ).copy()
+        if parent != root_id and parent != "":
```
ChrissiWHG: Is it really necessary that it is checked against ""? The root_id is a root identifier, or "" as you get it from ``_find_root_entity_id``.

If it is necessary:
Check against the root node in an explicit helper function or somehow different at a single place that it is easier to extend. Then the 2 different checks are more clear.

stable_plugins/classical_ml/data_preparation/one-hot_encoding.py
```diff
@@ -438,6 +457,9 @@ def prepare_stream_output(
                 sub_attributes.add(values)
 
             for sub_attribute in sub_attributes:
+                if sub_attribute == root_id or sub_attribute == "":
```
ChrissiWHG: Same for check against "" and root_id here.

ChrissiWHG (changes requested):

ChrissiWHG (approved):

PR #194, opened 2026-06-24
Consistent input data type handling

maxschuele:
Closes UST-QuAntiL/QHAna4MUSE-EnPro26#42


SebastianBruy:
Hey, I looked through the issue and the changes. As far as I understand it, the actual handling for JSON/CSV input formats is mostly implemented now.

I’m just not completely sure whether this is already enough, or whether a few things should still be added:

* Do the Wu-Palmer workflow/BPMN files also need to be updated, so they use the new plugin versions and content types?
* Are there already route/metadata tests, to check that the plugins actually expose the new supported input formats correctly?
* Would it make sense to move the JSON/CSV/JSON Lines loading logic into a shared helper function, so that each plugin does not contain similar loading logic separately or is it unnecessary?

maxschuele:
> Hey, I looked through the issue and the changes. As far as I understand it, the actual handling for JSON/CSV input formats is mostly implemented now.
> 
> I’m just not completely sure whether this is already enough, or whether a few things should still be added:
> 
>    - Do the Wu-Palmer workflow/BPMN files also need to be updated, so they use the new plugin versions and content types?

I updated the versions of the plugins in the bpmn files. No other changes necessary since it can just use the existing content types.
 
>  - Are there already route/metadata tests, to check that the plugins actually expose the new supported input formats correctly?

No, but i would say this is out of scope for this issue. Testing the celery tasks is sufficitent for now since no logic changed for the routes/schemas only additional data types were added. Can be addressed here: https://github.com/UST-QuAntiL/QHAna4MUSE-EnPro26/issues/56

>  - Would it make sense to move the JSON/CSV/JSON Lines loading logic into a shared helper function, so that each plugin does not contain similar loading logic separately or is it unnecessary?

The logic for this is already in shared helper functions (entity_marshaling.py) which are used here. The only special case is loading the entities from csv using the attribute metadata in the vector_concat and one_hot plugins which is needed due to multi valued attribute fields. The logic here was copied from the wu_palmer plugin.
We could add a way so this does not have to be done manually for example by passing the attribute metadata to the `load_entities` function. I will create a followup issue for this, needs to be discussed before implementation.

@SebastianBruy

SebastianBruy:
Thanks for the clarification.

PR #197, opened 2026-06-30
Feature/taxanomy mapping distances plugin

ChrissiWHG:


qhana_plugin_runner/__init__.py
```diff
@@ -244,6 +309,8 @@ def create_app(test_config: Optional[Dict[str, Any]] = None, silent_log: bool =
         # register jinja debug extension
         app.jinja_env.add_extension("jinja2.ext.debug")
 
+    download_mathjax(app)
```
buehlefs: Please move every change in this file, except this call, into its own python module! Possible locations: `plugin_runner/markdown.py`, `plugin_runner/mathjax.py`, `plugin_runner/util/mathjax.py`

Also: I am not sure if it is a good idea to execute this function on every start! This would also trigger in the test environment, which is not exactly desirable. (that is why I proposed adding it to the install plugin dependencies CLI function (<https://github.com/UST-QuAntiL/qhana-plugin-runner/blob/main/qhana_plugin_runner/plugins_cli.py>)

qhana_plugin_runner/__init__.py
```diff
@@ -50,12 +56,71 @@
...
+)
 
 
 def load_toml(file_like: IO[Any]) -> Mapping[str, Any]:
     return parse_toml("\n".join(file_like.readlines()))
 
 
+def validate_hash(response: requests.Response, expected_hash):
+    content = response.content
+    # (Achtung: Bei stream=True musst du aufpassen, response.content lädt alles in den Speicher.
```
buehlefs: please only use English comments to be consistent with the rest of the project.

qhana_plugin_runner/__init__.py
```diff
@@ -50,12 +56,71 @@
...
+MATHJAX_TGZ_HASH = (
+    "sha384-YX0aQfgsztN33UJH8sLOBH/+aJx5hhPpRupYKw3+G+L1EgqoC0/083Yy6omnOQs0"
+)
 
 
 def load_toml(file_like: IO[Any]) -> Mapping[str, Any]:
     return parse_toml("\n".join(file_like.readlines()))
 
 
+def validate_hash(response: requests.Response, expected_hash):
```
buehlefs: the input to this function should be the bytes content, not the request. Then this function would no longer need to worry about whether the request was streamed.

Also: Public/all functions should have doc-strings!

qhana_plugin_runner/__init__.py
```diff
@@ -50,12 +56,71 @@
...
+    mathjax_dir = Path(app.static_folder) / "mathjax"
+    marker = mathjax_dir / "es5" / "tex-mml-chtml.js"
+
+    if marker.exists():
+        return
+
+    mathjax_dir.mkdir(parents=True, exist_ok=True)
+
+    version = "3.2.2"
+    tarball_url = f"https://registry.npmjs.org/mathjax/-/mathjax-{version}.tgz"
```
buehlefs: This should at minimum be a CONSTANT like the hash at the top of the file.
Even better: Put all three config values into the app config (see `qhana_plugin_runner/util/config/__init__.py`) and read the app config here to get the actual values.

qhana_plugin_runner/__init__.py
```diff
@@ -50,12 +56,71 @@
...
+
+    response = requests.open_url(tarball_url, raise_on_error_status=True, stream=True)
```
buehlefs: requests.open_url can also be used with the `with` block for implicit resource cleanup.

qhana_plugin_runner/templates/simple_template.html
```diff
@@ -54,6 +54,12 @@
 
     {% block script %}
     <script src="{{url_for('static', filename='microfrontend.js')}}"></script>
+
+    <script>window.MathJax = { tex: { inlineMath: [["$", "$"]] } };</script>
+    <script src="{{ url_for('static', filename='mathjax/es5/tex-mml-chtml.js') }}"
+        integrity="sha384-Wuix6BuhrWbjDBs24bXrjf4ZQ5aFeFWBuKkFekO2t8xFU0iNaLQfp2K6/1Nxveei" crossorigin="anonymous"
```
buehlefs: The hash should also be read from config (i.e. from the same place where the mathjax version is configured).

qhana_plugin_runner/__init__.py
```diff
@@ -50,12 +56,71 @@
...
+                    target_path.mkdir(parents=True, exist_ok=True)
+                    continue
+
+                if member.isfile():
+                    target_path.parent.mkdir(parents=True, exist_ok=True)
+                    extracted = archive.extractfile(member)
+                    if extracted is None:
+                        continue
+                    with extracted, target_path.open("wb") as output_file:
+                        output_file.write(extracted.read())
```
buehlefs: pathlib path objects have a direct [write_bytes](https://docs.python.org/3/library/pathlib.html#pathlib.Path.write_bytes) method. (also a write_text one)

qhana_plugin_runner/__init__.py
```diff
@@ -50,12 +56,71 @@
...
+            package_prefix = "package/"
+            for member in archive.getmembers():
+                if not member.name.startswith(package_prefix):
+                    continue
+
+                relative_path = member.name[len(package_prefix) :]
+                if not relative_path or relative_path.startswith("."):
+                    continue
+
+                target_path = mathjax_dir / relative_path
```
buehlefs: please also test, that the target_path is indeed relative to the mathjax_dir! This would be a possible path traversal attack vector if not! (it is somehow mitigated by only allowing for a known tar file based on the hash, but better safe than sorry)

qhana_plugin_runner/__init__.py
```diff
@@ -50,12 +56,71 @@
...
+        raise ValueError(
+            f"Invalid hash. Expected: {expected_hash}, Actual: {actual_hash}"
+        )
+
+
+def download_mathjax(app, expected_hash=MATHJAX_TGZ_HASH):
+    mathjax_dir = Path(app.static_folder) / "mathjax"
+    marker = mathjax_dir / "es5" / "tex-mml-chtml.js"
+
+    if marker.exists():
```
buehlefs: If you also store the hash for this script in the app options (or as a constant here), then you could detect new versions and autoupdate the mathjax scripts.

stable_plugins/classical_ml/data_preparation/mapping_distances/tests/test_helpers.py
```diff
@@ -0,0 +1,227 @@
...
+        meta = _meta(ref_target="taxonomies.zip:color.json")
+        assert extract_tax_name(meta) == "color"
+
+    def test_ref_target_without_json_suffix(self):
+        meta = _meta(ref_target="taxonomies.zip:color")
+        assert extract_tax_name(meta) == "color"
+
+    def test_ref_target_nested_name(self):
+        meta = _meta(ref_target="taxonomies.zip:some/nested.json")
+        assert extract_tax_name(meta) == "some/nested"
```
buehlefs: Not sure if this case should be supported... My gut feeling is that these zips should be flat or rejected.

stable_plugins/classical_ml/data_preparation/mapping_distances/tests/test_helpers.py
```diff
@@ -0,0 +1,227 @@
...
+    def test_blank_string_value_returns_empty(self):
+        assert get_element_list({"color": "   "}, "color", _meta()) == []
+
+    def test_multiple_string_value_is_split_on_separator(self):
+        meta = _meta(multiple=True, separator=";")
+        result = get_element_list({"color": "red; blue ;green"}, "color", meta)
+        assert result == ["red", "blue", "green"]
+
+    def test_multiple_without_separator_is_treated_as_single(self):
+        meta = _meta(multiple=True, separator="")
```
buehlefs: Multiple True without a separator should probably raise a value error in the helper method building the deserializer... Better would be specifying another separator (e.g. $ or |)

stable_plugins/classical_ml/data_preparation/mapping_distances/__init__.py
```diff
@@ -0,0 +1,72 @@
...
+    __name__,
+    description="Taxanomy mapping to distances plugin API.",
+)
+
+
+class MappingDistances(QHAnaPluginBase):
+    name = _name
+    version = _version
+    description = (
+        "A plugin to create pairwise element distances for taxanomy mappings.<br /><br />\n"
```
buehlefs: For a new paragraph in markdown, ensure a blank line between text lines. Do not overuse manual line breaks with `<br />` ...
Also: for larger descriptions using triple quotes might be nicer.

stable_plugins/classical_ml/data_preparation/mapping_distances/routes.py
```diff
@@ -0,0 +1,163 @@
...
+            name=MappingDistances.instance.name,
+            version=MappingDistances.instance.version,
+            type=PluginType.processing,
+            entry_point=EntryPoint(
+                href=url_for(f"{MAPPING_DISTANCES_BLP.name}.{ProcessView.__name__}"),
+                ui_href=url_for(f"{MAPPING_DISTANCES_BLP.name}.{MicroFrontend.__name__}"),
+                data_input=[
+                    InputDataMetadata(
+                        data_type="entity/list",
+                        content_type=["application/json"],
```
buehlefs: this should support more than just json

stable_plugins/classical_ml/data_preparation/mapping_distances/routes.py
```diff
@@ -0,0 +1,163 @@
...
+                        required=True,
+                        parameter="entitiesUrl",
+                    ),
+                    InputDataMetadata(
+                        data_type="entity/attribute-metadata",
+                        content_type=["application/json"],
```
buehlefs: same here, as there is also a csv serialization if I remember

stable_plugins/classical_ml/data_preparation/mapping_distances/tasks.py
```diff
@@ -0,0 +1,273 @@
...
+    """Extracts the taxonomy name from the attribute metadata"""
+    tax_name = ""
+    if (
+        attrib_meta
+        and attrib_meta.ref_target
+        and "taxonomies.zip:" in attrib_meta.ref_target
+    ):
+        tax_name = attrib_meta.ref_target.split("taxonomies.zip:")[1]
+        if tax_name.endswith(".json"):
+            tax_name = tax_name[:-5]
```
buehlefs: This contains 2 very bad assumptions:
1. all taxonomy files will be called "taxonomies.zip". They could just as well be named "foo.bar". The important and useful bit here is the `:` (so split at the colon if a colon is present), also use the max_split param of the split function
2. Taxonomies will always be `.json`. While definitely true now, the better strategy here would be to use Pathlib to remove the file ending from the filename like to get the name only part.

stable_plugins/classical_ml/data_preparation/mapping_distances/tasks.py
```diff
@@ -0,0 +1,273 @@
...
+    val = entity.get(attribute)
+    if val is None:
+        return []
+    if isinstance(val, (set, list, dict)) and not val:
+        return []
+    if isinstance(val, (set, list)):
+        return [str(v) for v in val if v]
+    if isinstance(val, str):
+        if metadata.multiple and metadata.separator:
+            return [v.strip() for v in val.split(metadata.separator) if v.strip()]
```
buehlefs: You could use the inline assignment operator `:=` to avoid stripping the string twice, but that is not really required here.

stable_plugins/classical_ml/data_preparation/mapping_distances/tasks.py
```diff
@@ -0,0 +1,273 @@
...
+        return [val.strip()] if val.strip() else []
+    return [str(val)]
+
+
+def calculate_vector_distance(
+    v1: List[float], v2: List[float], metric: DistanceMetricEnum, db_id: int
+) -> float:
+    """
```
buehlefs, line 150: There is a vscode plugin that can help with generating docstrings from function signature

stable_plugins/classical_ml/data_preparation/mapping_distances/tasks.py
```diff
@@ -0,0 +1,273 @@
...
+    Calculates the distance between two coordinate vectors based on the selected metric.
+
+    Returns ``sys.float_info.max`` if the vectors are empty, i.e. no mapping is assigned.
+
+    Throws a ``ValueError`` if the mapping vectors do not have the same size.
+    """
+    if len(v1) != len(v2):
+        msg = f"Vectors do not have the same length in mapping_distances plugin task with db_id '{db_id}'"
+        TASK_LOGGER.error(msg)
```
buehlefs: It would be cleaner for this function to only throw the error and the parent/calling function with access to the db_id of the task to log the error.

stable_plugins/classical_ml/data_preparation/mapping_distances/tasks.py
```diff
@@ -0,0 +1,273 @@
...
+        raise ValueError(msg)
+
+    if not v1 or not v2:
```
buehlefs: Maybe also check for Nan values.

stable_plugins/classical_ml/data_preparation/mapping_distances/tasks.py
```diff
@@ -0,0 +1,273 @@
...
+            tax_map = taxonomy_mappings.get(tax_name, {})
+
+            unique_elements = set()
+            for entitiy in entities:
+                entity_names = get_element_list(entitiy, attribute, attrib_meta)
+                unique_elements.update(entity_names)
+
+            elements = sorted(list(unique_elements))
+            for e1 in elements:
+                for e2 in elements:
```
buehlefs: There is an itertool for this: <https://docs.python.org/3/library/itertools.html#itertools.product>

stable_plugins/classical_ml/data_preparation/mapping_distances/tasks.py
```diff
@@ -0,0 +1,273 @@
...
+                            "distance": dist,
+                        }
+                    )
+
+            with StringIO() as file:
+                save_entities(element_distances, file, "application/json")
+                file.seek(0)
+                zip_file.writestr(f"{attribute}.json", file.read())
+
+    zip_file.close()
```
buehlefs: should not be required as zip file is created within a with block

stable_plugins/muse/muse_for_music/__init__.py
buehlefs, line 1: Revert if this change was only for internal testing.

qhana_plugin_runner/templates/simple_template.html
```diff
@@ -54,6 +54,12 @@
 
     {% block script %}
     <script src="{{url_for('static', filename='microfrontend.js')}}"></script>
+
+    <script>window.MathJax = { tex: { inlineMath: [["$", "$"]] } };</script>
```
buehlefs: Also configure display math with $$ ... $$?

qhana_plugin_runner/templates/simple_template.html
```diff
@@ -54,6 +54,12 @@
...
+    <script src="{{ url_for('static', filename='mathjax/es5/tex-mml-chtml.js') }}"
```
buehlefs: Maybe use conditional loading: <https://docs.mathjax.org/en/latest/web/loading.html#loading-mathjax-only-on-pages-with-math>
(could also be an issue for another day if it is not changed in this PR)

buehlefs (changes requested):
There are many comments, but most of them should be easy to address (all comments with could/maybe or similar are optional)

qhana_plugin_runner/templates/simple_template.html
ChrissiWHG: $$ ... $$ already works, even without the commented line 58. 
And $ ... $ only with it. 
$$ ... $$ also does not work if the whole script is not included, which means that this setup works for inlineMath and separated line "$$" commands.

qhana_plugin_runner/templates/simple_template.html
ChrissiWHG: Das dort verlinkte snippet macht auch nur $ ... $.

stable_plugins/classical_ml/data_preparation/mapping_distances/tests/test_helpers.py
ChrissiWHG: Created a Ticket for that [#90 ](https://github.com/UST-QuAntiL/QHAna4MUSE-EnPro26/issues/90)

stable_plugins/classical_ml/data_preparation/mapping_distances/__init__.py
ChrissiWHG: Hm, with """ it does not work somehow. Both, the plugin description in the web and in the docs is in a code block or something strange. 
I added br and \n because it is rendered differently in the doc and in the web. This is the best I could achieve. Ending a line with "  \n" or just "\n" doen't change anything. Adding a "\n\n" also doesn't add an empty line. 

If you have a specific better string, please provide it.

stable_plugins/muse/muse_for_music/__init__.py
ChrissiWHG: Somehow the plugin is version v1.1.0 from others changes now. I reverted it, and have to re revert it haha.

ChrissiWHG: I don't know what happened. I am pretty sure that I saw a different version here. Maybe I had a specific commit opened?

stable_plugins/classical_ml/data_preparation/mapping_distances/__init__.py
buehlefs: With triple quotes you should not need \n, as all newlines inside triple quotes are preserved:

```python
a = "a\nb"

b = """a
b"""

a==b
```

In fact, triple quotes preserve every whitespace as is, which includes any indentation, which can mess with markdown rendering. Either don't indent text inside triple quotes or use textwrap.dedent to remove the indentation.

(I hope this actually addresses your problems with triple quotes. But there is no need to change this if you already found another way.)

qhana_plugin_runner/util/download_mathjax.py
```diff
@@ -0,0 +1,103 @@
...
+
+    _clear_old_mathjax_files(mathjax_dir)
+    mathjax_dir.mkdir(parents=True, exist_ok=True)
+
+    with requests.open_url(
+        tarball_url, raise_on_error_status=True, stream=True
+    ) as response:
+        is_sha512 = "sha512-" in expected_integrity
+        hasher = hashlib.sha512() if is_sha512 else hashlib.sha384()
+        prefix = "sha512-" if is_sha512 else "sha384-"
```
buehlefs:
```python
known_hashfunctions = {"sha512-": hashlib.sha512, "sha384-": hashlib.sha384}
prefix, hasher = None, None
for p, f in known_hashfunctions.items():
    if expected_integrity.startswith(p):  # use startswith, not in!
        prefix = p
        hasher = f()
if hasher is None:
    raise ValueError("unknon hash function")  # throw error on error case early!
```

stable_plugins/classical_ml/data_preparation/mapping_distances/tasks.py
```diff
@@ -120,27 +128,36 @@ def get_element_list(
...
+                stripped for v in val.split(metadata.separator) if (stripped := v.strip())
+            ]
         return [val.strip()] if val.strip() else []
     return [str(val)]
 
 
-def calculate_vector_distance(
-    v1: List[float], v2: List[float], metric: DistanceMetricEnum, db_id: int
+def _is_empty_or_nan(val):
+    if not val:
```
buehlefs: If val can be a number here, then `not val` will be true for `val==0`! (use type hints for the function!)

stable_plugins/classical_ml/data_preparation/mapping_distances/tasks.py
```diff
@@ -120,27 +128,36 @@ def get_element_list(
...
+        return True
+    return not val or isinstance(val, str) and val.strip().lower() == "nan"
```
buehlefs: Similarly, this does not test for an actual number type NaN value

qhana_plugin_runner/util/download_mathjax.py
```diff
@@ -0,0 +1,103 @@
...
+            for member in archive.getmembers():
+                if not member.name.startswith(package_prefix):
+                    continue
+                relative_path = member.name[len(package_prefix) :]
+                if not relative_path or relative_path.startswith("."):
+                    continue
+
+                target_path = mathjax_dir / relative_path
+                target_path.absolute().relative_to(
+                    mathjax_dir.absolute()
```
buehlefs: you might want to call absolute on this path early (and not every loop iteration)

qhana_plugin_runner/util/download_mathjax.py
```diff
@@ -0,0 +1,103 @@
...
+                )  # Path Traversal protection
```
buehlefs: How does this prevent anything if the target path is not changed and the result of the check is not used? Use if + `is_relative_to` to make the intention clearer. (or wrap this check in a function named `check_for_path_traversal` or similar) Also, the comment should make it clear that this will raise an exception on a path traversal

buehlefs (approved):
Looks good, just found some minor issues.

You could also decide to include the `check-for-tex.js` content directly in the template (or put it in the template folder) to avoid passing around the config variables from the template via a global object. But that is not really required.

qhana_plugin_runner/util/download_mathjax.py
```diff
@@ -60,9 +75,14 @@ def download_mathjax(app) -> None:
     with requests.open_url(
         tarball_url, raise_on_error_status=True, stream=True
     ) as response:
-        is_sha512 = "sha512-" in expected_integrity
-        hasher = hashlib.sha512() if is_sha512 else hashlib.sha384()
-        prefix = "sha512-" if is_sha512 else "sha384-"
+        known_hashfunctions = {"sha512-": hashlib.sha512, "sha384-": hashlib.sha384}
+        prefix, hasher = None, None
+        for p, f in known_hashfunctions.items():
+            if expected_integrity.startswith(p):  # use startswith, not in!
```
buehlefs: maybe remove the comment (was only intended for the review, not for the final code)

qhana_plugin_runner/util/download_mathjax.py
```diff
@@ -26,6 +26,21 @@ def _clear_old_mathjax_files(mathjax_dir: Path) -> None:
             path.unlink()
 
 
+def _verify_path_traversal(target_path: Path, abs_base_directory: Path) -> None:
```
buehlefs: rename to "_check_path_traversal". With verify it sounds like the path needs to be a path traversal attack to pass...

buehlefs (approved):
Looks good, only two nitpicks.

PR #198, opened 2026-07-03
Added new entity transformer plugin

Simonandy28:
# Descritption
Added new transformer plugin, than transforms element similarities to element distances. Instead of the "old" transformer plugin which transformed attribute similarities to attribute distances.

# Output Comparison
image


plugins/transformer/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+
+    tmp_zip_file = SpooledTemporaryFile(mode="wb")
+    zip_file = ZipFile(tmp_zip_file, "w")
+
+    for attribute in attributes:
+        element_distances = []
+        attribute_element_sims = element_similarities[attribute]
+
+        for sim_entity in attribute_element_sims:
+            sim = sim_entity.get("similarity")
```
maxschuele: use sim_entity["similarity"] like in the old transformer, should raise error if similarity key is missing

plugins/transformer/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+            dist = None
+
+            # Apply transformation algorithms
+            if sim is None:
+                dist = None
```
maxschuele: should raise a ValueError (also fix in old transformer plugin)

plugins/transformer/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+                beta = 1.0
+                dist = 1.0 / (1.0 + pow(sim / alpha, beta))
+            elif transformer == TransformersEnum.square_inverse:
+                max_sim = 1.0
+                dist = (1.0 / math.sqrt(2.0)) * math.sqrt(2.0 * max_sim - 2 * sim)
+
+            dist_entity = sim_entity.copy()
+
+            # Replace the 'similarity' key with 'distance'
+            if "similarity" in dist_entity:
```
maxschuele: redundant if statement, when changes i commented above are implemented

stable_plugins/classical_ml/data_preparation/transformers.py
```diff
@@ -54,15 +54,15 @@
 from qhana_plugin_runner.tasks import save_task_error, save_task_result
 from qhana_plugin_runner.util.plugins import QHAnaPluginBase, plugin_identifier
 
-_plugin_name = "sim-to-dist-transformers"
-__version__ = "v0.2.2"
+_plugin_name = "attr-sim-to-attr-dist-transformers"
+__version__ = "v0.3.1"
```
maxschuele: either 0.2.3 or 0.3

plugins/transformer/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+                dist = math.exp(-sim)
+            elif transformer == TransformersEnum.gaussian_inverse:
+                dist = math.exp(-sim * sim)
+            elif transformer == TransformersEnum.polynomial_inverse:
+                alpha = 1.0
+                beta = 1.0
+                dist = 1.0 / (1.0 + pow(sim / alpha, beta))
+            elif transformer == TransformersEnum.square_inverse:
+                max_sim = 1.0
+                dist = (1.0 / math.sqrt(2.0)) * math.sqrt(2.0 * max_sim - 2 * sim)
```
maxschuele: add else case that raises if transformer does not match any of the enum variants

plugins/transformer/schemas.py
```diff
@@ -0,0 +1,65 @@
...
+    )
+    attributes = ma.fields.String(
+        required=True,
+        allow_none=False,
+        metadata={
+            "label": "Elements",
+            "description": "Elements for which the similarity shall be transformed to distance.",
+            "input_type": "textarea",
+        },
+    )
```
maxschuele, line 51: Can attributes can be removed a parameter? We already selected the attributes we want in the wu-palmer step so the zip only contains the attributes we care about

maxschuele (changes requested):

plugins/transformer/tasks.py
```diff
@@ -0,0 +1,127 @@
...
+    zip_file.close()
+
+    filename = retrieve_filename(similarities_url)
+    info_str = f"_transformer_{transformer.name}_from_{filename}"
+
+    STORE.persist_task_result(
+        db_id,
+        tmp_zip_file,
+        f"transformers_elem_dist{info_str}.zip",
+        "custom/element-distances",
```
maxschuele: relation/element-distances

plugins/transformer/tests/test_tasks.py
```diff
@@ -0,0 +1,89 @@
...
+                assert "distance" in actual_pair
+
+                if expected_pair["distance"] is None:
+                    assert actual_pair["distance"] is None, f"pair {key(actual_pair)}"
+                else:
+                    assert actual_pair["distance"] == pytest.approx(
+                        expected_pair["distance"]
+                    ), f"pair {key(actual_pair)}"
+
+
```
maxschuele, line 83: add a test case to check that missing similarities raise an error

maxschuele (changes requested):

stable_plugins/classical_ml/data_preparation/transformers.py
```diff
@@ -54,15 +54,15 @@
 from qhana_plugin_runner.tasks import save_task_error, save_task_result
 from qhana_plugin_runner.util.plugins import QHAnaPluginBase, plugin_identifier
 
-_plugin_name = "sim-to-dist-transformers"
-__version__ = "v0.2.2"
+_plugin_name = "attr-sim-to-attr-dist-transformers"
```
maxschuele, line 57: Update references in bmpn files, other plugins and docs

plugins/transformer/tasks.py
Simonandy28: With the check "if sim is None:" an error is raised if .get("") is used.

PR #200, opened 2026-07-06
Aggregator plugin for attribute distances

maxschuele:
Closes UST-QuAntiL/QHAna4MUSE-EnPro26#76

### DIFF
--- a/plugins/aggregator/tasks.py
+++ b/plugins/aggregator/tasks.py
@@ -14,6 +14,8 @@
 
 import json
 from io import StringIO
+from itertools import combinations
+from pathlib import PurePath
 from tempfile import SpooledTemporaryFile
 from typing import Callable, Dict, List, Optional, Tuple
 from zipfile import ZipFile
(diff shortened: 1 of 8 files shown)


stable_plugins/classical_ml/data_preparation/aggregators.py
```diff
@@ -30,30 +29,29 @@
...
+from qhana_plugin_runner.requests import retrieve_filename
 from qhana_plugin_runner.storage import STORE
 from qhana_plugin_runner.tasks import save_task_error, save_task_result
 from qhana_plugin_runner.util.plugins import QHAnaPluginBase, plugin_identifier
-from qhana_plugin_runner.requests import retrieve_filename
-
 
-_plugin_name = "distance-aggregator"
-__version__ = "v0.2.2"
+_plugin_name = "entity-distance-aggregator"
```
SebastianBruy: Renaming the existing plugin from `distance-aggregator` to `entity-distance-aggregator` breaks existing references. For example, the EnPro router and the existing Wu-Palmer BPMN workflow still request `distance-aggregator` (including version `v0.2.2`).

Since the new plugin already has the distinct name `attribute-distance-aggregator`, could we keep the existing plugin name and version unchanged in this PR? Otherwise, all runtime references, workflows, tests, and documentation need to be migrated together.

plugins/aggregator/tasks.py
```diff
@@ -0,0 +1,227 @@
...
+
+
+def _load_entities(entities_url: str) -> List[dict]:
+    with open_url(entities_url) as entities_data:
+        mimetype = get_mimetype(entities_data)
+        entities = []
+        deserializer: Callable[[tuple[str, ...]], tuple[any, ...]] | None = None
+        attribute_metadata: dict[str, AttributeMetadata] = {}
+
+        attribute_metadata_url = entities_data.headers["X-Attribute-Metadata"]
```
SebastianBruy: This makes `X-Attribute-Metadata` mandatory for every entity input. Valid JSON and JSON-Lines entity files without this optional header will fail with a `KeyError`, even though their values are already deserialized.

Could we check whether the header exists before loading the metadata, similar to the previous Sym Max Mean implementation? A test with JSON input without `X-Attribute-Metadata` should also be added.

SebastianBruy (changes requested):
The core attribute-distance aggregation and the accompanying tests look reasonable. However, I found two issues that should be addressed before merging:

1. Renaming the existing `distance-aggregator` plugin breaks current router and BPMN workflow references.
2. Entity inputs without the optional `X-Attribute-Metadata` header fail with a `KeyError`, including otherwise valid JSON and JSON-Lines inputs.

Please address these issues before merging.

stable_plugins/classical_ml/data_preparation/aggregators.py
maxschuele: Reverted to the old name and version number. We can decide next sprint meeting how/if to rename the old plugins.

plugins/aggregator/tasks.py
maxschuele: Changed it so it only raises a error if entities mimetype is csv and the header is missing since the metadata is required to deserialize entities from csv.

SebastianBruy (approved):
The changes look good to me.

PR #203, opened 2026-07-12
MDS attribute distances plugin

maxschuele:
Closes UST-QuAntiL/QHAna4MUSE-EnPro26#77

### DIFF
--- a/plugins/attribute_mds/tasks.py
+++ b/plugins/attribute_mds/tasks.py
@@ -124,8 +124,8 @@ def calculation_task(self, db_id: int) -> str:
     filename = retrieve_filename(params.attribute_distances_url)
     filenames_hash = get_readable_hash(filename)
     info_str = (
-        f"_mds_dim_{params.dimensions}_metric_{metric_name}"
-        f"_from_{filename}_{filenames_hash}"
+        f"_mds_dim_{params.dimensions}_{metric_name}"
+        f"_{params.missing_data_handling.name}_from_{filename}_{filenames_hash}"
     )
 
     tmp_zip_file = SpooledTemporaryFile(mode="wb")
--- a/plugins/attribute_mds/tasks.py
+++ b/plugins/attribute_mds/tasks.py
@@ -76,7 +76,9 @@ def _replace_missing_distances(
             dist["distance"] = replacement
 
 
-def _build_distance_matrix(distances: List[dict]) -> Tuple[Dict[str, int], Any]:
+def _build_distance_matrix(
+    distances: List[dict], attr_name: str
+) -> Tuple[Dict[str, int], Any]:
     import numpy as np
 
     id_to_idx: Dict[str, int] = {}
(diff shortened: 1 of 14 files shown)


plugins/attribute_mds/tasks.py
```diff
@@ -0,0 +1,172 @@
...
+
+        mds = manifold.MDS(
+            params.dimensions,
+            metric=params.metric == MetricEnum.metric_mds,
+            n_init=params.n_init,
+            max_iter=params.max_iter,
+            dissimilarity="precomputed",
+        )
+        transformed = mds.fit_transform(distance_matrix)
+
```
SebastianBruy, line 189: Non-metric MDS in scikit-learn treats off-diagonal dissimilarities with value `0` as missing values. However, zero is a valid and common attribute distance when two entities have identical values for an attribute.

This means that selecting `nonmetric_mds` can silently ignore valid distances, and an attribute containing only zero distances may fail entirely. Could we either remove the non-metric option, handle off-diagonal zero values before calling scikit-learn, or reject this case with a clear validation error?

Please also add a test for non-metric MDS with valid zero distances.

plugins/attribute_mds/tasks.py
```diff
@@ -0,0 +1,172 @@
...
+    import numpy as np
+
+    id_to_idx: Dict[str, int] = {}
+
+    for dist in distances:
+        for ent_id in (dist["source"], dist["target"]):
+            if ent_id not in id_to_idx:
+                id_to_idx[ent_id] = len(id_to_idx)
+
+    distance_matrix = np.zeros((len(id_to_idx), len(id_to_idx)))
```
SebastianBruy: Initializing the complete matrix with zeros means that an omitted entity pair is silently interpreted as having distance `0`. This differs from an explicitly provided `null` value, although both represent missing input data.

Could we validate that every unordered entity pair occurs exactly once, or initialize missing entries separately and process them through the selected missing-data handling? A regression test with one omitted pair would be useful.

SebastianBruy (changes requested):
The general implementation matches the goal of producing one MDS embedding per attribute, and the metric MDS scenarios are covered well.

I found one functional issue with non-metric MDS: scikit-learn interprets valid zero distances as missing values, but zero distances can occur naturally in the attribute-distance pipeline. This option therefore needs additional handling or validation and a corresponding regression test.

I also recommend validating that the input contains every required entity pair, since omitted pairs are currently initialized as distance zero.

maxschuele:
Good points you raised here @SebastianBruy, i just adapted the code from the old mds plugin so these issues have to be addressed there too.

Will have to think about what is the best way to handle these.

plugins/attribute_mds/tasks.py
maxschuele: I decided to raise an error on missing pairs as this would indicate invalid data from the previous (aggregator) plugin whereas a null distance value for a pair is the case that one of the entities does not have a value for the attribute (missing data handling fills these then).

Also added a test case for this.

plugins/attribute_mds/tasks.py
maxschuele: Options for this:
- Remove nonmetric mds
- Raise an error if nonmetric mds is selected and a distance is zero
- Assign a small value to zero distances

buehlefs: From the [documentation](https://scikit-learn.org/stable/modules/manifold.html#multi-dimensional-scaling-mds):

> In the non-metric version, the algorithm will try to preserve the order of the distances, and hence seek for a monotonic relationship between the distances in the embedded space and the input dissimilarities.

This sounds like an option we should try to keep. Since it only tries to preserve the order between distances (i.e., if d_a > d_b in the input then this will not flip to < in the output) then using a small Value in place of 0 would make sense. However, this needs to be documented clearly in the user interface (and the plugin description). If possible, we could replace 0 with 0.000001 (or something similar) while building the matrix from the data (or while reading in the data) if non-metric is selected.

maxschuele:
For nonmetric MDS zero distancses are now replaced with 0.000001, added information about this to the plugin descriptions and added unit tests.

Also fixed the issues in the old MDS implementations and added tests there.

SebastianBruy:
Thanks, the missing-pair validation and the additional tests look good.

There is still one scale-dependent issue with using the fixed value `0.000001`. If the input contains a positive distance smaller than that value, replacing zero with `0.000001` changes the ordering of the dissimilarities.

For example:

* `d(A, B) = 0`
* `d(A, C) = 0.0000005`

After replacement, `d(A, B)` becomes larger than `d(A, C)`, although it was originally smaller. This conflicts with the goal of non-metric MDS to preserve the distance ordering.

Could the replacement instead be derived from the smallest positive off-diagonal distance, ensuring that `0 < replacement < smallest_positive_distance`, with a separate fallback for the case where all off-diagonal distances are zero?

Please also add a test containing a positive distance below `0.000001`. The same adjustment would be needed in the old MDS implementation.

maxschuele:
@SebastianBruy implemented replacement as `0 < replacement < smallest_positive_distance` and added tests.

SebastianBruy:
Thanks, so this looks good to me.

PR #205, opened 2026-07-14
Router plugin updates

Simonandy28:
- [x] Added checkbox to include intermediate results in output (closes UST-QuAntiL/QHAna4MUSE-EnPro26#73)
- [x] Added "None" as default setting for entitiy selection in substep
- [x] Replaced old plugins (Removed SymMaxMean, Added new Transformer, Aggregator and MDS) (closes UST-QuAntiL/QHAna4MUSE-EnPro26#78)
- [x] Updated Tests


plugins/enpro26/router/__init__.py
```diff
@@ -20,7 +20,7 @@
 from qhana_plugin_runner.util.plugins import QHAnaPluginBase, plugin_identifier
 
 _plugin_name = "router"
-__version__ = "v0.1.1"
+__version__ = "v0.1.3"
```
ChrissiWHG, line 23: Why the jump? It is not "released" yet. I think it could stay at v0.1.0 or does any other plugin depend on the higher of this plugin.

ChrissiWHG (changes requested):

plugins/enpro26/router/__init__.py
buehlefs: Any change in the plugin metadata requires a change in the plugin version (otherwise the plugin registry will ignore that change). Because of this it is better to be generous with version bumps (at least with minor version increases).

ChrissiWHG: Okay. Locally, I normally restart the engine. But increasing the version is also valid then.

plugins/enpro26/router/tasks.py
```diff
@@ -223,86 +219,43 @@ def handle_webhook_task(self, db_id: int, source_url: str):
...
 @CELERY.task(
-    name=f"{Router.instance.identifier}.process_step_3_transformers",
+    name=f"{Router.instance.identifier}.process_step_2_transformers",
     bind=True,
     max_retries=10,
 )
-def process_step_3_transformers(self, db_id: int, source_url: str):
-    TASK_LOGGER.info("Starting Step 3: Transformer")
+def process_step_2_transformers(self, db_id: int, source_url: str):
+    TASK_LOGGER.info("Starting Step 2: Transformer")
```
ChrissiWHG: Not that important, but the name is in plural, transformers.

plugins/enpro26/router/routes.py
```diff
@@ -315,6 +309,8 @@ def post(self, arguments, db_id: int):
         wu_palmer_attributes = [
             attr for attr, option in selections.items() if option == "Wu-Palmer"
         ]
+        # TODO: Add other paths (OneHot, NumMapping)
+
         unsupported = {
             attr: option for attr, option in selections.items() if option != "Wu-Palmer"
```
ChrissiWHG: It shouldn't treat "None" as unsupported. None should be skipped?

Or change message that it is not skipped due to the not yet implemented pipeline, but because none is selected.

plugins/enpro26/router/tasks.py
ChrissiWHG: So here is also should be "transformers" with an s in this string. Which means that the PIPELINE_PLUGINS field is named wrongly.

plugins/enpro26/router/tests/test_tasks.py
```diff
@@ -132,18 +160,21 @@ def mock_get(url, **kwargs):
...
-    def mock_delay(*args, **kwargs):
-        triggered.append("step_2_triggered")
+    def mock_apply_async(*args, **kwargs):
+        # We also want to assert that countdown=4 is actually being passed
+        if kwargs.get("countdown") == 4:
+            triggered.append("next_step_triggered")
 
     monkeypatch.setattr("requests.get", mock_get)
-    monkeypatch.setattr("router.tasks.process_step_2_smm.delay", mock_delay)
+    # We now mock apply_async instead of delay
```
ChrissiWHG: Remove comment

ChrissiWHG (changes requested):
I am not sure if the naming of the methods in `tasks.py` is a good idea. When adding the mapping to distances plugin and one hot encoding, it will change. 
But this could also be a future problem.
Another idea is to extract some methods (eg. helper functions) to different files before it gets more (the file is already large), but currently it is sufficiently readable.

ChrissiWHG (approved):

PR #206, opened 2026-07-20
Fix plugin identifiers whitespace

maxschuele:
Closes https://github.com/UST-QuAntiL/QHAna4MUSE-EnPro26/issues/49


plugins/circuit_executors/braket_local.py
```diff
@@ -59,7 +59,7 @@
 from qhana_plugin_runner.tasks import save_task_error, save_task_result
 from qhana_plugin_runner.util.plugins import QHAnaPluginBase, plugin_identifier
 
-_plugin_name = "AmazonBraket_LocalSimulator"
+_plugin_name = "amazon-braket-local-simulator"
```
ChrissiWHG: isn't the name ``amazonBraket_localSimulator`` also valid? Why the change. 

But if you want to match the convention that CamelCase is not used, we also have "_", so maybe the related words could be written together like ``amazon_bracket-local_simulator``.

plugins/circuit_executors/braket_local.py
```diff
@@ -59,7 +59,7 @@
...
 __version__ = "v1.0.0"
```
ChrissiWHG: Isn't that a version increase if the name is changed?

qhana_plugin_runner/util/plugins.py
```diff
@@ -23,6 +25,40 @@
...
+
+def _is_valid_plugin_name(name: str) -> bool:
+    """Check that a plugin name starts with a lowercase letter and contains only
+    letters, digits, hyphens, and underscores."""
+    return bool(_PLUGIN_NAME_REGEX.fullmatch(name))
+
+
+_PLUGIN_VERSION_REGEX = re.compile(
+    r"""
+    v?
```
ChrissiWHG: My search in the repo says that there is never a version number without v. Why do you want to accept that? I think this should be removed completely. Just unnecessary complexity.

If there are plugins without v, I think we should add a v instead of accepting it.

tests/test_plugin_names.py
```diff
@@ -0,0 +1,167 @@
...
+    Registration is the only public API that validates plugin names and
+    versions, so the checks go through a throwaway plugin class. The global
+    plugin registry is restored afterwards to keep the probe invisible to
+    other tests.
+    """
+    plugins = QHAnaPluginBase.get_plugins()
+    identifier = plugin_identifier(name, version)
+    registered_before = plugins.get(identifier)
+
+    probe = type("_ProbePlugin", (QHAnaPluginBase,), {"name": name, "version": version})
```
ChrissiWHG: Is that the german word Probe?

tests/test_plugin_names.py
```diff
@@ -0,0 +1,167 @@
...
+        plugins[identifier] = registered_before
+    elif accepted:
+        plugins.pop(identifier, None)
+    return accepted
+
+
+@pytest.mark.parametrize(
+    "name",
+    [
+        "a",
```
ChrissiWHG, line 61: I find it useful to test a variation of valid and invalid names and version numbers, but the more important thing is to check that the actual plugins are named correctly. You probably hit every wrong naming here, but we can't be sure that there is no version with all 0 or without v or a wrong name. 

So the test only validates that the two regexes in plugin.py are correct and not that the plugins are named correctly.

ChrissiWHG (changes requested):

tests/test_plugin_names.py
maxschuele: Created a followup issue for this: https://github.com/UST-QuAntiL/QHAna4MUSE-EnPro26/issues/104

Not sure which of the two approaches described in the issue is the better one -> discuss in next sprint meeting

tests/test_plugin_names.py
maxschuele: No the english word probe. Maybe mock would be less confusing, will change.

plugins/circuit_executors/braket_local.py
maxschuele: will change to `amazonBraket_localSimulator`, camel case is fine as long as the name regex fulfilled

ChrissiWHG (approved):

PR #207, opened 2026-07-22
Add mapping plugin to router

Simonandy28:
- [x] Added mapping plugin to router (closes UST-QuAntiL/QHAna4MUSE-EnPro26#93)
- [x] Refactored router tasks file structure
- [x] Added recomendation marker to ui for attribute selection
- [x] Updated Tests


plugins/enpro26/router/tasks.py
```diff
@@ -141,30 +67,51 @@ def preprocessing_task(self, db_id: int) -> str:
...
+            if ref:
+                tax_filename = PurePath(ref).name
+                # Ensure the taxonomy file exists in the zip
+                matched_zip_path = next((name for name in available_taxonomies if name.endswith(tax_filename)), None)
+                
+                if matched_zip_path:
+                    taxonomy_attributes.append(metadata.ID)
+                    
+                    # Open the taxonomy JSON and check for mappings
+                    try:
```
ChrissiWHG: you could outsource the recomendation calculation to a different method. It is confusing that you start checking the json for mappings here.

Then, the TODO of refining recommendations detection can be resolved easier. Is that TODO documented in a ticket or will it be solved in this issue?

plugins/enpro26/router/tasks.py
```diff
@@ -174,226 +121,52 @@ def preprocessing_task(self, db_id: int) -> str:
 @CELERY.task(name=f"{Router.instance.identifier}.route_task", bind=True)
 def route_task(self, db_id: int) -> str:
```
ChrissiWHG: I was confused on what the task does and why it is there at all. So I had to look up what happens before and after.

I would recommend to add a doc comment that it is called if the selection in the second UI step is done and the pipeline is already set up. The name `route_task` could be changed to `start_routing`, but I am not sure if that's better. I am bad at naming "talking" method names.

Another question: Why is there so much logic in the `routes.py`? The logic that happens before starting this task looks good but I thought the pattern is to have only task start and save boilerplate in the routes and everything else in tasks.  
Is it possible to move that here?

plugins/enpro26/router/tasks_pipeline_steps.py
```diff
@@ -0,0 +1,307 @@
+# Copyright 2026 QHAna plugin runner contributors.
```
ChrissiWHG, line 1: Maybe rename file that is closer to `tasks.py` -> tasks_pipeline_steps eg. Not sure about that name,

plugins/enpro26/router/routes.py
```diff
@@ -309,28 +311,42 @@ def post(self, arguments, db_id: int):
...
-            if option != "Wu-Palmer" and option != "None"
-        }
-        if unsupported:
-            db_task.add_task_log_entry(
-                f"Pipelines not implemented yet, attributes skipped: {unsupported}"
-            )
-        db_task.data["wu_palmer_attributes"] = "\n".join(wu_palmer_attributes)
+        
+        db_task.data["pipeline_queue"] = pipeline_queue
+        db_task.data["current_pipeline"] = None
```
ChrissiWHG: Move the attribute extraction and pipeline building to a task if that is possible as suggested in the other comment. The db_task is also accessible there, so I think it should work.

plugins/enpro26/router/routes.py
ChrissiWHG, line 324: I am not sure if that can be moved. Not completely sure what adding these urls does later and where it should be located.

plugins/enpro26/router/routes.py
ChrissiWHG: You do not use the task_data parameters before calling the route_task. So they are not needed before line 367, I think?

plugins/enpro26/router/pipeline_tasks.py
```diff
@@ -0,0 +1,307 @@
...
+        subscribe_to_plugin(task_url, task_data.data["webhook_url"])
+    except (requests.exceptions.ReadTimeout, requests.exceptions.ConnectionError) as e:
+        task_data.add_task_log_entry(f"Error occured during wu-palmer plugin. Attempting to retry:\n{e}")
+        task_data.save(commit=True)
+        raise self.retry(exc=e, countdown=CELERY_COUNTDOWN)
+    except Exception as e:
+        task_data.add_task_log_entry(f"CRASH in wu_palmer step:\n{traceback.format_exc()}")
+        save_task_error.delay(failing_task_id=self.request.id, db_id=db_id)
+
+# MAPPING TASK
```
ChrissiWHG: Comment is different without ---

plugins/enpro26/router/pipeline_tasks.py
```diff
@@ -0,0 +1,307 @@
...
+    if next_pipeline == "wu_palmer":
+        task_data.add_task_log_entry("Starting Wu-Palmer Pipeline. Includes: Wu-Palmer, Transformer, Aggregator, MDS")
+        task_data.save(commit=True)
+        start_wu_palmer.apply_async(args=[task_data.id])
+    elif next_pipeline == "mapping":
+        task_data.add_task_log_entry("Starting Distance Mapping Pipeline. Includes: Distance Mapping, Aggregator, MDS")
+        task_data.save(commit=True)
+        start_mapping.apply_async(args=[task_data.id])
+
+# WU-PALMER TASK
```
ChrissiWHG: Comment is different without ---

plugins/enpro26/router/schemas.py
```diff
@@ -137,6 +149,25 @@ class InputParametersSchema(FrontendFormBaseSchema):
         metadata={"label": "Root is part of hierarchy", "input_type": "checkbox"},
     )
 
+    distance_metric = EnumField(
```
ChrissiWHG, line 185: The distance metric is now above the transformer in the UI.  
But why? Mapping doesn't come before Transformer or after it.

Is there a way to make sections in the UI. It became very large and gets progressively larger.

If you know something about that or have any idea, we can make an issue for that. It is not necessary to do it here.

plugins/enpro26/router/schemas.py
ChrissiWHG, line 230: A first idea to make the UI clearer is is to adapt the labels of the plugins that it is clearer what you are setting up. 

Metric is not very clear to what it belongs. Now, we also have the Distance Metric.

ChrissiWHG (changes requested):
As we already talked about that, some comments could be removed or improved. 

But in general I really like the much more compact `tasks.py` with seperation of interest.

plugins/enpro26/router/pipeline_tasks.py
```diff
@@ -0,0 +1,307 @@
...
+        task_data.save(commit=True)
+
+        subscribe_to_plugin(task_url, task_data.data["webhook_url"])
+    except (requests.exceptions.ReadTimeout, requests.exceptions.ConnectionError) as e:
+        task_data.add_task_log_entry(f"Error occured during wu-palmer plugin. Attempting to retry:\n{e}")
+        task_data.save(commit=True)
+        raise self.retry(exc=e, countdown=CELERY_COUNTDOWN)
+    except Exception as e:
+        task_data.add_task_log_entry(f"CRASH in wu_palmer step:\n{traceback.format_exc()}")
+        save_task_error.delay(failing_task_id=self.request.id, db_id=db_id)
```
ChrissiWHG, line 97: If I understand correctly, every method does exactly the same, even the exception handling when it once defined the payload. So this could be outsourced to a single method getting the payload and the plugin name. This should be sufficient to get the url and write all logging messages.

plugins/enpro26/router/pipeline_tasks.py
```diff
@@ -0,0 +1,307 @@
...
+        response.raise_for_status()
+
+        task_url = urljoin(wu_palmer_url, response.headers["Location"])
+        task_data.data["wu_palmer_url"] = task_url
+        task_data.add_task_log_entry("Started Wu-Palmer.")
+        task_data.save(commit=True)
+
+        subscribe_to_plugin(task_url, task_data.data["webhook_url"])
+    except (requests.exceptions.ReadTimeout, requests.exceptions.ConnectionError) as e:
+        task_data.add_task_log_entry(f"Error occured during wu-palmer plugin. Attempting to retry:\n{e}")
```
ChrissiWHG: Why is the error message after 'Attempting to retry:'. Are we retrying the error? Please rewrite the message.

ChrissiWHG (changes requested):

plugins/enpro26/router/tasks.py
Simonandy28: I will write a new issue for it. It will not be solved in this issue.

plugins/enpro26/router/tasks_pipeline_steps.py
Simonandy28: I think tasks_pipeline_steps is good

plugins/enpro26/router/schemas.py
Simonandy28: I do not know whether you can add sections to the UI. 
I also think the idea was to make some of the settings adjustable for each attribute specific. 
Eitherway the input screen will need more changes in the future, but not in this pr. 
I will make an issue and adjust the descriptions in the ui to make it more understandable for now.

plugins/enpro26/router/tasks_pipeline_steps.py
```diff
@@ -26,17 +26,62 @@
...
 from .schemas import InputParameters, InputParametersSchema
-from .task_helpers import subscribe_to_plugin, extract_output_url, _plugin_process_url
+from .tasks_helpers import subscribe_to_plugin, extract_output_url, _plugin_process_url
 
 TASK_LOGGER = get_task_logger(__name__)
 CELERY_COUNTDOWN = 3
 
 # Collection of celery tasks for starting plugins that are substeps of the various pipelines managed by the routing plugin.
 
+# --- BOILERPLATE HELPERS ---
```
ChrissiWHG: How we have helper methods in the extra file that separates code from helpers.

These helpers probably should go into the `tasks_helpers.py` file.

plugins/enpro26/router/tasks_pipeline_steps.py
```diff
@@ -26,17 +26,62 @@
...
+    task_data.add_task_log_entry(f"Started {step_name}.")
+    task_data.save(commit=True)
+
+    subscribe_to_plugin(task_url, task_data.data["webhook_url"])
+
+
+# --- PIPELINE ORCHESTRATION ---
+
+
 def launch_next_pipeline(task_data: ProcessingTask):
```
ChrissiWHG, line 42: I would say that it is up to you whether this function is a helper or not. Maybe directly in `tasks.py` is also an option.

All other functions in this file are Celery tasks.

plugins/enpro26/router/tasks_pipeline_steps.py
```diff
@@ -26,17 +26,62 @@
...
 
 TASK_LOGGER = get_task_logger(__name__)
 CELERY_COUNTDOWN = 3
 
 # Collection of celery tasks for starting plugins that are substeps of the various pipelines managed by the routing plugin.
 
+# --- BOILERPLATE HELPERS ---
+
+
+def run_pipeline_step(celery_task, db_id: int, step_log_name: str, step_action: callable):
```
ChrissiWHG: I don't want to be tedious, but this function does not run the pipeline step. It is only error handling for creating the data that is handled to `invoke_plugin_and_subscribe` INSIDE the `step_action`. 

This is a bit confusing on what method does what. 

I have two ideas:
- pass all parameters from `invoke_plugin_and_subscribe` to this method and actually start it here. Then the other method could be removed, unless you want a separate error handling.
- Just rename the methods and make clear what is happening.

ChrissiWHG: I don't know how python works, but you are passing `self` as the first parameter. Why is it called `celery_task` and not self?

Maybe it doesn't work or did you think that is more clear with that name? Maybe this can be cleared by refactoring in general.

ChrissiWHG: You also wouldn't need the callable if you simply pass everything to this method.

plugins/enpro26/router/tasks_helpers.py
```diff
@@ -101,3 +104,23 @@ def _load_entity_attributes(entities_url: str) -> set:
                 break
             attributes.update(ent.keys())
     return attributes
+
+
+def _calculate_recommendations(taxonomies_zip: ZipFile, zip_path: str) -> str:
```
ChrissiWHG: Underscore is used to annotate private functions. Again, I don't know that much about Python, but my IDE says that these private functions are never called, but they are actually called in other files.

So maybe rename them to match the naming convention.

plugins/enpro26/router/tasks.py
```diff
@@ -118,55 +117,116 @@ def preprocessing_task(self, db_id: int) -> str:
...
-    
+
     current_pipeline = task_data.data.get("current_pipeline")
 
     if current_pipeline == "wu_palmer":
         handle_wu_palmer_progression(task_data, db_id, source_url)
-        
+
     elif current_pipeline == "mapping":
         pass
```
ChrissiWHG: I think the pass should be removed. I am surprised that it even works. That is Python magic...

plugins/enpro26/router/tasks_pipeline_steps.py
Simonandy28: I kept them there because they are very specific for the sub tasks. But `task_helpers`  is also fine.

plugins/enpro26/router/tasks_pipeline_steps.py
Simonandy28: Moving it to tasks.py would cause a circular import because finalize_pipeline uses launch_next_pipeline.
And I would rather not move it to helpers. 
I think its fine there

plugins/enpro26/router/tasks_pipeline_steps.py
```diff
@@ -222,19 +208,30 @@ def action(task_data):
...
         )
+        save_task_error.delay(failing_task_id=self.request.id, db_id=db_id)
+        return
 
-    run_pipeline_step(self, db_id, "Aggregator", action)
+    run_pipeline_step(
+        celery_task=self,
+        db_id=db_id,
+        task_data=task_data,
+        plugin_name="aggregator",
```
ChrissiWHG: Just one small thing...
This plugin is named aggregators and then the `url_key` parameter can be removed, because it is always plugin_name_url. And it should always be correct to have it like that I think.

ChrissiWHG (changes requested):

ChrissiWHG (approved):

PR #208, opened 2026-07-22
Feature/add zip to vector concat

ChrissiWHG:
Added ZIP handling to the vector concat plugin


ChrissiWHG:
Closes UST-QuAntiL/QHAna4MUSE-EnPro26#97

qhana_plugin_runner/requests.py
```diff
@@ -130,7 +130,7 @@ def get_mimetype(response: Response, default=None) -> Optional[str]:
         return response.headers["Content-Type"]
     except KeyError:
         matches = mimetypes.MimeTypes().guess_type(url=response.url)
-        if matches:
+        if matches and matches[0]:
```
ChrissiWHG, line 133: I suppose this was a bug and I changed if for every place, because actually matches is never undefined, because it is a tuple.

Simonandy28 (approved):
Looks good.

PR #209, opened 2026-07-28
Add vector concat to router

ChrissiWHG:
I didn't add tests for the tasks, because I want to rebuild the error handling and wanted to think about the tests later. Created an issue for that: UST-QuAntiL/QHAna4MUSE-EnPro26#100

PR #210, opened 2026-08-03
Improve error handling of router

ChrissiWHG:
Closes https://github.com/UST-QuAntiL/QHAna4MUSE-EnPro26#94

Has to be reviewed after https://github.com/UST-QuAntiL/qhana-plugin-runner/pull/209

### DIFF
--- a/plugins/enpro26/router/routes.py
+++ b/plugins/enpro26/router/routes.py
@@ -365,11 +365,11 @@ class WebhookView(MethodView):
     def post(self, db_id: int):
         source_url = request.args.get("source")
         event = request.args.get("event")
+        via = request.args.get("via")  # "watchdog" if delivered by the polling watchdog
 
         if source_url and event == "status":
-            # Countdown prevent overload for backend.
             handle_webhook_task.apply_async(
-                kwargs={"db_id": db_id, "source_url": source_url}, countdown=2
+                kwargs={"db_id": db_id, "source_url": source_url, "via": via},
+                countdown=2,
             )
-
         return "Webhook received", HTTPStatus.OK
--- a/plugins/enpro26/router/routes.py
+++ b/plugins/enpro26/router/routes.py
@@ -359,7 +359,12 @@ class RoutingStepView(MethodView):
 # --- WEBHOOOK ---
 @ROUTER_BLP.route("/<int:db_id>/webhook/")
 class WebhookView(MethodView):
-    """Endpoint to receive webhook updates from called plugins."""
+    """
+    Receives and processes state update webhooks from executed sub-plugins.
+
+    Listens for 'status' events from plugins currently executing a pipeline step.
+    It triggers the `handle_webhook_task` to progress the next step.
+    """
 
     @ROUTER_BLP.response(HTTPStatus.OK)
     def post(self, db_id: int):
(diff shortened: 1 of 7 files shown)


ChrissiWHG:
Filename handling of vector concat is changed here, but still buggy. This will be solved in a separate ticket: https://github.com/UST-QuAntiL/QHAna4MUSE-EnPro26/issues/103

Simonandy28:
Bug: Tasks keep on running, but ui registered failure:

Complete Log that shows that the tasks keep on going:
[CompleteLog.log](https://github.com/user-attachments/files/30856482/CompleteLog.log)

Log from the UI:
[uiLog.log](https://github.com/user-attachments/files/30856483/uiLog.log)


image

ChrissiWHG:
This last commit closes https://github.com/UST-QuAntiL/QHAna4MUSE-EnPro26/issues/103.

ChrissiWHG:
I never got this issue because my machine is probably faster. So it could be that some problems only occur if the worker is more in overload.

The error occurred, because the Celery.TASK class child PipelineTask cannot access the current flask app in the on_retry hook.

After fixing it, I tested it by setting `REQUEST_TIMEOUT = 1` to simulate an easier heavy workload.
Then, a ReadTimeoutError occurred multiple times, but it was consequently logged in the UI: [UILog_working.log](https://github.com/user-attachments/files/30863320/UILog_working.log) and in the console:
grafik
I uploaded an image, because then it is highlighted and we can see that the watchdog actually got triggered at the end. Unfortunately the logging of why, when and how the watchdog executed a lost task, because something in the subscription went wrong, is not there. And I didn't have time to investigate it.
This is also not displayed in the UI. This could be improved.

Accidentally, in the middle of the execution with some errors, I ended the worker with `Ctrl+C` and after starting the worker again, it finished successfully.

ChrissiWHG:
Okay now I took a closer look on the logging and the normal task got three times a ReadTimeout error. Then, the watchdog won the race of finishing first and we got the logging in the `handle_webhook_task` that it finished with watchdog. Some loggings later, there is the message, that the normal task finally finished:
`[2026-08-08 22:54:07,241: INFO/MainProcess] Task router@v0-1-5.handle_webhook_task[145b6724-8e60-42d3-b2ed-905155eac4cd] succeeded in 0.009018762000778224s: 'Sub-task already progressed'`
But then it was already processed by the watchdog. 

So there is no problem. 
This happened multiple times after three tried executions or more specifically at recognizing that it is finished and the `CELERY_COUNTDOWN` is set to 3 for "using" the monitor/watchdog.

ChrissiWHG:
I added one commit in the PR https://github.com/UST-QuAntiL/qhana-plugin-runner/pull/215 that is linked here.
I didn't want to push it directly here, because it should only be used to think of how to change the current behavior and how to build useful logging.

This logging shows that the watchdog is useless if it doesn't try 3 times and actually use it. I also think that the execution was slower, but I am not sure.

This is my test output in the UI:
[UILog_watchdog.log](https://github.com/user-attachments/files/30863992/UILog_watchdog.log)

ChrissiWHG:
@Simonandy28 you can continue on this or wait until I am back.

PR #215, opened 2026-08-08
add extensive logging, also to ui to analyze watchdog usage

ChrissiWHG:
Didn't want to commit that in the actual branch because this solution is not good but gives good insights. See https://github.com/UST-QuAntiL/qhana-plugin-runner/pull/210 for explanation.


Simonandy28 (approved):
I would remove logs from the ui in the future and add them to the console. 
For now its fine-

